PR #12, opened 2025-04-29
Fix cactus dependency path

leo-gall:
This PR changes the path to the `cactus_flutter` dependency. Before this PR, the Flutter example application wasn't found and `flutter pub get` threw the following error:
```
Resolving dependencies... 
Because cactus_flutter_example depends on cactus_flutter from path which doesn't exist (could not find package cactus_flutter at "../flutter"), version solving failed.
Failed to update packages.
```

I recommend merging #11 before merging this PR.


HenryNdubuaku:
@leo-gall I will get back to this soon, sorry for the delayed approval, i made some changes in the folder and will push the Flutter soon, will let you know :)

PR #51, opened 2025-05-24
Sbb gh patch 1

sbb-gh:


cactus/ggml-cpu/ggml-cpu.c
HenryNdubuaku, line 1: @sbb-gh thanks! any specific reason for adding modifying this particular file? The other changes are perfect!

Issue #29, opened 2025-05-04
GitHub Commit Workflows


HMUNACHI:
Set up GitHub workflows to run tests and generate docs after every PR

a-ayomide:
@HenryNdubuaku @rshemet Hi team, can I work on this?

HenryNdubuaku:
Hey Ayo, thanks for following up. Please proceed! Are you in the Discord?

a-ayomide:
Yes I am. I'll proceed on this.

a-ayomide:
Hi team (@HenryNdubuaku @rshemet), I couldn't find the tests for the modules but I added the gha to setup and generate docs for the react module. This lays the foundation for adding other ghas as well. Will also appreciate a pointer to the tests for the other modules incase I missed it. PR can be found here --> (https://github.com/cactus-compute/cactus/pull/95)

HenryNdubuaku:
Thanks for this! this is fine, I'll review the PR. Tests are tricky because our engine is designed for phones and need to run in an emulator. So testing is restricted to the CPP

PR #95, opened 2025-07-18
ci: added gha for running tests in react dir and generating docs

a-ayomide:


a-ayomide:
REF: https://github.com/cactus-compute/cactus/issues/29

PR #117, opened 2025-08-11
Unify Model Initialization 96

Ok3ks:
I know I possibly jumped the gun, by starting this PR without asking here. 

This PR tracks current progress of this branch. 

- currently takes local path 
- there's a thrown `cactusException` when model file does not exist
- Example updated with an additional `downloadModel` called before `CactusVLM.init()`

Is there a deprecation guide to follow, I simply removed these parameters for respective functions.


devabhixda:
I'm working to split download and init, should help sync the flows across different platform libs

devabhixda:
Completed the changes https://github.com/cactus-compute/cactus/pull/120

PR #161, opened 2025-10-13
Update engine.h with smol model type, Update engine with smol configuration and tokenizer

justinl66:
Dependent on PR #160


HenryNdubuaku:
@justinl66 I like this PR, both INT8 and FP16 work fine! Simply pull, fix the merge conflicts, and push, so I'll merge.

PR #165, opened 2025-10-14
Fixed SentencePiece tokenizer

ncylich:
Small bug where pre-processed texts wouldn't usually start with a "▁". Fixed this and also filtered out leading spaces in line with hf's implementation.


HenryNdubuaku:
Thanks for noticing! @ncylich run git pull first and test on gemma, remember to setup your DCO, so I can merge.

HenryNdubuaku:
@ncylich Tested and merged

PR #166, opened 2025-10-16
warn user if unsaved tensors exist

ammesatyajit:
added code that tracks unsaved tensors and warns the user. initially had it raise an error, but the qwen models had one weight that wasn't being saved.


HenryNdubuaku:
@ammesatyajit I think you still need to run tools/setup-dco.sh, this will automatically sign off your commits.

PR #171, opened 2025-10-16
Removed layernorm op from graph and implemented using existing ones instead + Optimized scatter_topk for long sequence lengths

ncylich:


tests/test_engine.cpp
HenryNdubuaku, line 1: This only hides underlying error

PR #179, opened 2025-10-21
fixed attention kernel compile error

kar-m:


kar-m:
Wasn't compiling on debian, problem with a variable being float32x4 instead of int32x4. Compiled and tested the change

PR #191, opened 2025-10-30
Add relevant stb_image header files

justinl66:


HenryNdubuaku (dismissed):
Is this ready?

justinl66:
Now it is!

justinl66:
Overview of changes:
Kept image headers as 2 separate files. For various reasons like certain duplicate names and not wanting to deal with that, but essentially no downside as we are keeping them in a separate lib folder anyway. (The combined total is around 11k lines, the 8k lines I had earlier was likely due to accidentally deleting too much and therefore the lib didn't compile.)
Emitting header definitions in an engine_image.c file instead of C++. This is because the files are meant to be compiled in a C99 standard while cactus make actually uses cpp. To combat this, I compiled the headers as C files then included into engine.h using extern "C".

HenryNdubuaku:
@justinl66 is the engine_image implementation C? given the .c extension?

justinl66:
the stb image headers are written for C/C++, but stb_image_resize2 specifically is using features that only C99 has. So yes, engine_image is emitting the headers in C, but the headers are written to be compatible with C++. It's not a problem (unless it's something to do with optimization that I don't know about) but it compiles cleanly under the makefile.

PR #199, opened 2025-11-04
Fix premature stopping of text generation

devabhixda:
**The Problem:**

The stop sequence matching was tokenizing stop sequences like "<|im_end|>" into multiple tokens, then adding each individual token to a stop set. This caused generation to halt prematurely when any token from a stop sequence appeared anywhere in the output.

**The Fix:** Replaced the broken token-based matching with proper sequence matching:

- Created matches_stop_sequence() function that checks if the entire sequence of tokens appears at the end of generated tokens
- Updated cactus_complete() to build stop_token_sequences instead of individual token sets
- Now correctly continues generation until a complete stop sequence is encountered

### DIFF
--- a/cactus/ffi/cactus_ffi.cpp
+++ b/cactus/ffi/cactus_ffi.cpp
@@ -382,9 +382,7 @@ int cactus_complete(
 
         generated_tokens.push_back(next_token);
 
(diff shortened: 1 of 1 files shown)


HenryNdubuaku:
This looks goo @devabhixda I am merging now!

PR #210, opened 2025-11-09
Add multiple tool call test to engine tests

rshemet:
Updated the test_engine.cpp file to include a new test function, test_tool_call_with_multiple_tools, which verifies the functionality of multiple tools in a single call. Also modified the model path for consistency with the latest weights.

### DIFF
--- a/tests/test_engine.cpp
+++ b/tests/test_engine.cpp
@@ -7,7 +7,7 @@
 #include <vector>
 #include <sstream>
 
-const char* g_model_path = "../../weights/lfm2-350m";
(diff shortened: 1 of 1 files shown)


tests/test_engine.cpp
```diff
@@ -7,7 +7,7 @@
 #include <vector>
 #include <sstream>
 
-const char* g_model_path = "../../weights/lfm2-1.2b";
```
HenryNdubuaku, line 10: Thanks, this changes the default model we use, git pull the latest and merge.

PR #212, opened 2025-11-15
Justin/lfm2 rag prod

justinl66:


tests/test_engine.cpp
```diff
@@ -330,6 +330,26 @@ bool test_huge_context() {
...
 }
 
+bool test_rag() {
+    const char* messages = R"([
+        {"role": "system", "content": "You are a helpful assistant."},
+        {"role": "user", "content": "What has Justin been doing at Cactus Candy?"}
+    ])";
+
+    std::string modelPathStr(g_model_path ? g_model_path : "");
+    if (modelPathStr.find("rag") == std::string::npos) {
```
HenryNdubuaku: this depends on the weight file name, but users can name their files anything, you wanna search for this in the config.txt since model variant has that information.

cactus/engine/engine_tokenizer.cpp
```diff
@@ -149,6 +172,49 @@ std::string Tokenizer::format_lfm2_style(const std::vector<ChatMessage>& message
         sys_content += "{\"function_call\": {\"name\": \"function_name\", \"arguments\": {\"arg1\": \"value1\"}}}";
     }
 
+    if (model_variant_ == ModelVariant::RAG) {
```
HenryNdubuaku, line 175: this is hard coded to read from the the test files, so this cannot be used, the idea is to simply accept a document path in the FFi which is an optional param.

cactus/engine/engine_tokenizer.cpp
```diff
@@ -149,6 +172,49 @@ std::string Tokenizer::format_lfm2_style(const std::vector<ChatMessage>& message
...
+                    }
+                }
+            }
+            closedir(dir);
+
+            std::sort(files.begin(), files.end());
+
+            int doc_idx = 1;
+            for (const auto& fname : files) {
+                if (doc_idx > 1000) break;
```
HenryNdubuaku: The hard coded doc limit should be removed, users should be able to add a long doc if they want

tests/test_engine.cpp
```diff
@@ -7,7 +7,7 @@
 #include <vector>
 #include <sstream>
 
-const char* g_model_path = "../../weights/lfm2-1.2B";
+const char* g_model_path = "../../weights/lfm2-1.2b-rag-i8";
```
HenryNdubuaku: You wanna avoid changing default test settings in PRs else it breaks for people

HenryNdubuaku (commented):
This works! Just left a few comments

cactus/ffi/cactus_ffi.h
HenryNdubuaku, line 1: please no additional endpoint, simply add corpus dir as an optional parameter

tools/convert_hf.py
```diff
@@ -530,6 +530,7 @@ def _cfg_get(c, key, default=None):
         'use_layout_tags': bool(use_layout_tags),
         'tie_word_embeddings': tie_word_embeddings
     }
+    model_config['model_variant'] = 'vlm'
```
HenryNdubuaku: wann remove this

cactus/ffi/cactus_ffi.cpp
```diff
@@ -70,6 +70,45 @@ cactus_model_t cactus_init(const char* model_path, size_t context_size) {
     }
 }
 
+cactus_model_t cactus_init_with_corpus(const char* model_path, size_t context_size, const char* corpus_dir) {
```
HenryNdubuaku: No entirely new init function, simpy update the existign init with an optional parmameter

HenryNdubuaku (commented):
@justinl66 thanks! i have added more comments and a comment was unattended previously

PR #215, opened 2025-11-18
Refactor CLI build scripts into unified command-line tool

ncylich:
Replace platform-specific build scripts (build.sh/build.bat) with a new unified CLI tool that provides separate commands for building, downloading models, and running the chat interface. The new 'cactus' executable supports multiple model configurations and eliminates hardcoded model paths.

Key improvements:
- Single entry point with 'build', 'download', and 'run' subcommands
- Dynamic model selection via command-line arguments
- Automatic model directory name resolution from HuggingFace IDs
- Colored terminal output for better user feedback
- Consistent cross-platform behavior (Unix/Windows)
- Improved syntax in cactus_ffi.cpp

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ Example response from Gemma3-270m-INT8
 - NPUs to improve energy-efficiency and prefill speed up to 11x
 - VLM and Audio models like LFM-VL, Whisper, KittenTTS, etc. 
 
-## Using this repo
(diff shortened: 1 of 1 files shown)


HenryNdubuaku:
this is cool, move the unified cactus cli to the root, so users can just run `cactus run <model/name>`. Also, you wanna clean up the comments

PR #227, opened 2025-11-26
Instead of manually installing python deps and creating weights, use `requirements.txt` with the help of venv.

rozeappletree:
With this PR, we do not need to manually install deps for running `python3 tools/convert_hf.py LiquidAI/LFM2-VL-1.6B weights/lfm2-vl-1.6b/ --precision INT8`. 

- Works as long as python3 is available in dev env.
- Increases automation, reduces manual intervention

**Note:** 

This PR uses `.venv` instead of `venv` which is present in `.gitignore` file, change based on internal standards.

https://github.com/rozeappletree/cactus/blob/c1e87ed213cab3e49752659dfcec69d504293f98/tests/run.sh#L21

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ apple/*.xcframework
 apple/libcactus.a
 *.a
 venv
+.venv


HenryNdubuaku:
Thanks for this @rozeappletree so cli/cactus test command automatically sets up the venv for the user, have you reviewed that?

rozeappletree:
> Thanks for this @rozeappletree so cli/cactus test command automatically sets up the venv for the user, have you reviewed that?

I was running tests using `chmod +x tests/run.sh && ./tests/run.sh`

But the recent changes are way better than what is implemented in this PR.

closing as this is obsolete now.

PR #231, opened 2025-11-26
Add weight generation and release pipeline

devabhixda:
- Add GitHub Actions workflow to convert and upload models
- Configure model matrix with 14 supported models
- Implement Supabase storage upload for model weights
- Add metadata insertion to cactus schema
- Configure disk space management for CI
- Set up automatic zip archive creation
- Add release notes generation

### DIFF
--- a/.github/workflows/release-models.yml
+++ b/.github/workflows/release-models.yml
@@ -1,0 +1,252 @@
+name: Release Models to Supabase
+
+on:
+  push:
+    tags:
+      - 'v*'
+      - '*'
+
+jobs:
+  convert-and-upload-models:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        model:
+          - 'google/gemma-3-270m-it'
+          - 'LiquidAI/LFM2-350M'
+          - 'HuggingFaceTB/SmolLM2-360m-Instruct'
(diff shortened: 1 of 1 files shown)


.github/workflows/release-models.yml
```diff
@@ -0,0 +1,252 @@
...
+    steps:
+      - name: Create release notes
+        run: |
+          TAG_NAME=${GITHUB_REF#refs/tags/}
+          echo "## Model Release $TAG_NAME" >> release-notes.md
+          echo "" >> release-notes.md
+          echo "All models have been converted and uploaded to Supabase Storage." >> release-notes.md
+          echo "" >> release-notes.md
+          echo "Models included:" >> release-notes.md
+          echo "- google/gemma-3-270m-it" >> release-notes.md
```
HenryNdubuaku, line 238: Thanks for this Abhi! We need to include the whisper models

.github/workflows/release-models.yml
```diff
@@ -0,0 +1,252 @@
...
+          echo "- LiquidAI/LFM2-350M" >> release-notes.md
```
HenryNdubuaku, line 239: also, is the workflow run after every merge or release?

PR #236, opened 2025-11-27
fix(vlm): Resolve undefined variable in VLM conversion

harshaljanjani:
→ The `convert_hf_model_weights_vlm` function attempts to add tensor names to the `saved_tensor_full_names` set without initializing it first.


harshaljanjani:
@HenryNdubuaku, are the PRs garbled again?

PR #239, opened 2025-11-28
Optimized softmax to use Horner's method

ParkiratS:
Replaced existing approximation code to utilize Horner's method.


HenryNdubuaku:
How much performance gains do we get?

PR #246, opened 2025-12-10
Add telemetry framework

devabhixda:
Screenshot 2025-12-10 at 9 36 47 AM


cactus/ffi/cactus_complete.cpp
```diff
@@ -174,9 +175,28 @@ int cactus_complete(
         }
 
         std::strcpy(response_buffer, result.c_str());
+
```
HenryNdubuaku, line 178: Should be called CactusTelemetry

cactus/ffi/cactus_telemetry.h
```diff
@@ -0,0 +1,41 @@
+#ifndef CACTUS_TELEMETRY_H
```
HenryNdubuaku, line 1: we dont create separate headerfiles for everything unless its a header only libs.

cactus/ffi/cactus_telemetry.cpp
```diff
@@ -0,0 +1,344 @@
+#include <thread>
```
HenryNdubuaku: actually, this should be a header-only file; meaning that both the declarations and implementations should be in a unified cactus_telemetary.h, intead of a separate .cpp file

cactus/ffi/cactus_telemetry.cpp
```diff
@@ -0,0 +1,344 @@
...
+    bool enabled_ = false;
+    std::string telemetry_token_;
+    std::string project_id_;
+    std::string device_id_;
+
+    mutable std::mutex mutex_;
+};
+
+static const std::string SUPABASE_URL = "https://vlqqczxwyaodtcdmdmlw.supabase.co";
+static const std::string SUPABASE_KEY = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdXBhYmFzZSIsInJlZiI6InZscXFjenh3eWFvZHRjZG1kbWx3Iiwicm9sZSI6ImFub24iLCJpYXQiOjE3NTE1MTg2MzIsImV4cCI6MjA2NzA5NDYzMn0.nBzqGuK9j6RZ6mOPWU2boAC_5H9XDs-fPpo5P3WZYbI";
```
HenryNdubuaku: is this safe?

cactus/ffi/cactus_utils.h
```diff
@@ -15,12 +15,18 @@
...
+#ifdef __APPLE__
+#include <uuid/uuid.h>
+#endif
 
 struct CactusModelHandle {
     std::unique_ptr<cactus::engine::Model> model;
     std::atomic<bool> should_stop;
     std::vector<uint32_t> processed_tokens;
     std::mutex model_mutex;
+    std::string model_name;
```
HenryNdubuaku, line 29: we dont use this anywhere, its just gonna pass empty string as name to the telemetry

cactus/ffi/device_manager.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+#ifdef __APPLE__
+#include <sys/utsname.h>
+#include <unistd.h>
+#endif
+
+namespace cactus {
+namespace ffi {
+
+class HttpClient {
```
HenryNdubuaku: i think this is duplicated somewhere else now

cactus/ffi/device_manager.cpp
```diff
@@ -0,0 +1,254 @@
+#include "cactus_utils.h"
```
HenryNdubuaku: we should move the codes to cactus_telemetry

cactus/ffi/http_client.cpp
```diff
@@ -0,0 +1,104 @@
+#ifdef __APPLE__
```
HenryNdubuaku: move to cactus telemetry too, no need for separate files for these

VERSION
```diff
@@ -0,0 +1 @@
+1.2.1
```
HenryNdubuaku: version is just 1.2

cactus/ffi/cactus_utils.h
devabhixda: We are extracting model name in init: https://github.com/cactus-compute/cactus/pull/246/files/f4d9e59e38798cb37faeac5895e1a687d52338a0#diff-ade68cd762cba105c7f0eb81d691fe6dde7a2b384d88d07e547e05470555eb5dR31

cactus/ffi/cactus_telemetry.cpp
devabhixda: Yes it's anon key, we expose it in frameworks as well

cactus/ffi/device_manager.cpp
devabhixda: Done

PR #285, opened 2026-01-21
Stop generation on model end token

Ayan9074:
## Description
Ensures models stop generating once their end-of-turn token is reached. This prevents cases where models emit a stop token but continue generating additional text afterward. The stop token is suppressed from the final output by default, while still allowing users to override stopping behavior via custom stop words if needed.

## Type of Change
- [X] Bug fix
- [X] New feature
- [ ] Performance improvement
- [ ] Documentation update

## Testing
- [ ] Tests pass locally (VLM MULTI-TURN TEST fails with & without changes)
- [X] Tested on ARM hardware
- [ ] Benchmarked performance impact

## Checklist
- [X] All commits are signed-off (DCO)
- [X] Code follows project style
- [X] Comments added where necessary
- [X] Documentation updated if needed

### DIFF
--- a/cactus/engine/engine.h
+++ b/cactus/engine/engine.h
@@ -152,6 +152,7 @@ public:
     virtual uint32_t get_bos_token() const = 0;
     virtual uint32_t get_eos_token() const = 0;
     virtual bool has_chat_template() const { return has_chat_template_; }
+    std::string get_default_stop_sequence() const;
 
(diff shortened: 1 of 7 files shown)


HenryNdubuaku:
@Ayan9074 thanks for this! @rshemet wanna test this fix to see if you have any more problems

Issue #111, opened 2025-08-05
Insert media from buffer to loadPrompt()


ffreality:
Current `loadPrompt()` excepts medias as paths but technologies like game engines and camera apps have buffers rather than media paths. Writing actual files are performance consuming operation. Is there any way to pass it ?

I can extract raw BGRA color data and convert them RGB buffer. I looked mtmd and it excepts RGB. Also these buffers doesn't have any headers like jpeg or png. As I say before they are pure color array.

I couldn't saw any sample about that.

sudhanshu112233shukla:
@HenryNdubuaku @rshemet , @ffreality 
add this feature here! do check the PR

PR #287, opened 2026-01-23
Feature/raw image buffer

sudhanshu112233shukla:
#111
I added support for raw image buffers by extending the system across three core layers: the FFI (Public API), the Engine (Core Logic), and the Model (Implementation). The goal at the FFI layer was to allow external languages such as Python, Swift, and C to pass raw memory pointers instead of being restricted to image file paths. To achieve this, in ffi/cactus_ffi.h I introduced a new C struct called cactus_image_t that holds raw pixel data (void* data), image dimensions (width, height), and the number of channels. In ffi/cactus_complete.cpp, I implemented a new API function named cactus_complete_with_images. This function parses the input JSON messages, looks for special image identifiers such as "buffer://image1", matches those identifiers against the array of cactus_image_t structs provided by the caller, and packages all valid matches into a new internal structure called MediaInput.

At the engine layer, the goal was to introduce a unified abstraction capable of representing either an image loaded from disk or one stored entirely in memory, so that downstream components would not need to care about the image’s origin. For this, in engine/engine.h I introduced the MediaInput struct, which uses a tagged union approach with enum Type { PATH, BUFFER }. This design allows the engine to safely carry either a file path or a raw memory pointer down to the model layer without performing expensive memory copies at this stage, preserving performance and keeping the abstraction clean.

At the model layer, the objective was to teach the Vision Language Model (Lfm2VlModel) how to consume the new MediaInput abstraction. In models/model_lfm2vl.cpp, I created a new central method called decode_with_media. This method iterates through all inputs and, when it encounters a PATH, it invokes preprocess_from_file as before, while for a BUFFER it calls preprocess_from_memory. A crucial safety step here is that preprocess_from_memory, which already existed in Siglip2Preprocessor, immediately copies and normalizes the raw image bytes into the engine’s internal float tensors. This ensures that even if the caller frees the original raw buffer immediately after invoking the API, the engine remains safe because it owns its fully processed copy of the data.

To verify that the entire pipeline works correctly, I added a standalone test in tests/test_buffer.cpp. This test simulates a real user integration by manually creating a 2×2 pixel red image in C++ memory, wrapping it inside a cactus_image_t, and passing it to the new cactus_complete_with_images API. The test asserts that the model initializes correctly and accepts the in-memory image data without crashing. Together, this complete pipeline enables streaming video frames or in-memory generated images directly into the inference engine with zero reliance on the file system, while maintaining safety, performance, and architectural clarity.

### DIFF
--- a/cactus/engine/engine.h
+++ b/cactus/engine/engine.h
@@ -516,6 +516,7 @@ public:
     virtual uint32_t decode_with_media(const std::vector<uint32_t>& tokens, const std::vector<MediaInput>& media,
                                           float temperature = -1.0f, float top_p = -1.0f,
                                           size_t top_k = 0, const std::string& profile_file = "", float* out_entropy = nullptr) {
+        (void)tokens; (void)media; (void)temperature; (void)top_p; (void)top_k; (void)profile_file; (void)out_entropy;
         // Default implementation throws, similar to decode_with_images
         throw std::runtime_error("decode_with_media not implemented for this model");
     }
--- a/cactus/models/model_lfm2vl.cpp
+++ b/cactus/models/model_lfm2vl.cpp
@@ -463,7 +463,7 @@ Lfm2VlModel::ForwardImageResult Lfm2VlModel::forward_images(
         }
         gb->set_input(embedding_input.input_node, segment_data.data(), Precision::FP32);
     }
-    return language_model_.forward(gb, merged_embeddings.node_id, merged_embeddings.seq_len, backend, use_cache);
+    size_t final_hidden = language_model_.forward(gb, merged_embeddings.node_id, merged_embeddings.seq_len, backend, use_cache);
     return ForwardImageResult{final_hidden, merged_embeddings.seq_len};
 }
 


sudhanshu112233shukla:
@HenryNdubuaku @rshemet  do check IT

HenryNdubuaku:
Thanks for this @sudhanshu112233shukla, currently waiting for review from @Ayan9074

sudhanshu112233shukla:
@HenryNdubuaku always here to contribute , working on some more issues will open the PR and fix them soon!!

sudhanshu112233shukla:
> Hi, thanks for the PR. It currently doesn’t compile.
> 
> Errors received: In model_lfm2vl.cpp, forward_images() returns a size_t but the function signature expects a ForwardImageResult, and final_hidden is undefined. This causes the build to fail.
> 
> There are also a few unused-parameter warnings from the default decode_with_media implementation in engine.h.
> 
> Please fix the build error and clean up the warnings, and I’ll be happy to re-review.

just fixing it , will update it within an hr

sudhanshu112233shukla:
Thanks for the review, @Ayan9074! 
I’ve fixed the build errors and cleaned up the warnings:
Fixed model_lfm2vl.cpp: Updated forward_images to return the ForwardImageResult struct instead of size_t, which resolves the “undefined final_hidden” build error.
Cleaned up warnings: Added void casts for intentionally unused parameters in decode_with_media inside engine.h.
The build should be passing now. Ready for a re-review!

Issue #110, opened 2025-08-05
getFormattedChat() throws exception with target:image


ffreality:
Hi guys, you can see my VLM related codes for Unreal Engine 5 at bottom.

I init Cactus, its parameters and model paths like this. I checked and they successfully initialized with all required parameters.

<img width="1693" height="319" alt="Image" src="https://github.com/user-attachments/assets/913fcaaf-a18c-4982-be03-c10ee15361cb" />

**But `FormattedPrompt = this->Cactus_Context->getFormattedChat(Messages, "");` throws exeption like this. (you will see it at the bottom ACactus_Manager_VLM::GenerateResponseToImage)**

```
chat.cpp:127

catch (const std::exception & e) {
        // @ngxson : disable otherwise it's bloating the API response
        // printf("%s\n", std::string("; messages = ") + messages.dump(2));
        throw std::runtime_error("Failed to parse messages: " + std::string(e.what()));
}
```

I know that I didn't set any media paths but even if I remove everything besides `getFormattedChat()`, error continues. If I don't include `target : image` there is no problem.

My C++ code
```
bool ACactus_Manager_VLM::Init_Cactus(FCactusModelParams_VLM VLM_Params)
{
	if (Cactus_Context.IsValid())
	{
		UE_LOG(LogTemp, Warning, TEXT("Cactus Context is already initialized !"));
		return false;
	}

	if (!VLM_Params.bIsNumbersOkay())
	{
		UE_LOG(LogTemp, Warning, TEXT("Numeric values should bigger than 0 !"));
		return false;
	}

	if (this->Path_Model.IsEmpty())
	{
		UE_LOG(LogTemp, Warning, TEXT("Model path is not set !"));
		return false;
	}

	try
	{
		this->Cactus_Params.model.path = TCHAR_TO_UTF8(*this->Path_Model);
		this->Cactus_Params.n_ctx = VLM_Params.ContextSize;
		this->Cactus_Params.n_batch = VLM_Params.BatchSize;
		this->Cactus_Params.n_gpu_layers = VLM_Params.GPULayers;
		this->Cactus_Params.cpuparams.n_threads = VLM_Params.CPUThreads;

		this->Cactus_Context = MakeShared<cactus_context_vlm, ESPMode::ThreadSafe>();

		if (!this->Cactus_Context->loadModel(this->Cactus_Params))
		{
			UE_LOG(LogTemp, Error, TEXT("Failed to load Cactus model from path: %s"), *this->Path_Model);
[...]

PR #288, opened 2026-01-24
Fix Issue #110: Robust multi-modal message parsing for VLM support

sudhanshu112233shukla:
This PR fixes Issue #110 by improving how we parse multi-modal messages in cactus_utils.h.

The main problem was that modern SDKs (OpenAI / Anthropic) send content as an array of parts, not just a plain string. Because of this, VLM models like LFM2-VL were failing with an “Unsupported content part type” error. This change makes the parser fully compatible with those formats while keeping existing behavior intact.
What changed

Updated parse_messages_json to accept content as either a string or an array of objects

Added support for the following content part types:
- text
- image_url
- image
Improved key detection logic to avoid edge cases where values look like keys
(e.g. correctly handling "type": "text" vs the "text" field itself)

Preserved backward compatibility with the legacy images array format
Verification

Tested the changes using a Python test suite (ported from the C++ logic) across multiple formats:

Plain string content
{"role": "user", "content": "hello"} 

Multi-modal array content
{"role": "user", "content": [{"type": "text", ...}, {"type": "image_url", ...}]} 

Mixed / legacy format
{"role": "user", "content": "...", "images": ["..."]} 

Example
The library now officially supports multi-modal input like:
`[
  {
    "role": "user",
    "content": [
      {
        "type": "text",
        "text": "What is in this image?"
      },
      {
        "type": "image_url",
        "image_url": {
          "url": "/path/to/image.jpg"
        }
      }
    ]
  }
]
`

### DIFF
--- a/cactus/ffi/cactus_utils.h
+++ b/cactus/ffi/cactus_utils.h
@@ -176,35 +176,141 @@ inline std::vector<cactus::engine::ChatMessage> parse_messages_json(const std::s
         }
 
         size_t role_pos = json.find("\"role\"", pos);
-        if (role_pos == std::string::npos || role_pos >= obj_end) break;
+        if (role_pos != std::string::npos && role_pos < obj_end) {
+            size_t role_start = json.find('"', role_pos + 6) + 1;
+            size_t role_end = json.find('"', role_start);
+            msg.role = json.substr(role_start, role_end - role_start);
+        }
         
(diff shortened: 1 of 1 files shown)


sudhanshu112233shukla:
@HenryNdubuaku @rshemet  do check this PR hope this give it good fix to the issue , if it needs any further changes do tell me from my side i verified and tested it well ...

PR #290, opened 2026-01-26
Fix UTF-8 token callback issue for multi-byte characters (issue #90)

### LINKED ISSUE #90: Token callback error: Unfinished UTF-8 octet sequence
I’m using a Korean language model, but I’m getting the following error 

```
I/flutter ( 7509): 안
I/flutter ( 7509): 녕
I/flutter ( 7509): 히
I/flutter ( 7509):  주
I/flutter ( 7509): Token callback error: FormatException: Unfinished UTF-8 octet sequence (at offset 1)
E/flutter ( 7509): [ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: FormatException: Unfinished UTF-8 octet sequence (at offset 93)
```

Thanks!

sudhanshu112233shukla:
Fix UTF-8 Token Callback for Multi-Byte Characters (Fixes #90)
Fixes:#90
This fix solves a crash that happened when using languages like Korean, Chinese, Japanese, or emojis during token streaming. The issue occurred because individual tokens were decoded and sent to the callback immediately. Many of these languages use multi-byte UTF-8 characters, which can be split across multiple tokens. When an incomplete character was sent, it caused a FormatException: Unfinished UTF-8 octet sequence, leading to crashes.

To fix this, tokens are now buffered instead of being sent one by one. A small helper function checks whether the buffered text forms a complete UTF-8 character. The callback is triggered only when the UTF-8 sequence is complete. Any remaining buffered text is flushed at the end of generation to ensure nothing is lost.

The change was applied to both text completion and audio transcription callbacks. ASCII text continues to work exactly the same, while multi-byte characters like Korean, Chinese, Japanese, and emojis now stream smoothly without errors. The performance impact is negligible since the UTF-8 check only looks at the last few bytes.

After this fix, token callbacks no longer crash, multi-byte characters display correctly, and streaming output feels smooth and stable. There are no API changes this is a safe bug fix that restores proper UTF-8 handling.

### DIFF
--- a/cactus/ffi/cactus_complete.cpp
+++ b/cactus/ffi/cactus_complete.cpp
@@ -9,6 +9,43 @@ using namespace cactus::ffi;
 
 static constexpr size_t ROLLING_ENTROPY_WINDOW = 10;
 
+// Check if a string ends with a complete UTF-8 sequence
+static bool is_complete_utf8(const std::string& str) {
+    if (str.empty()) return true;
+    
+    // Check the last few bytes for incomplete UTF-8 sequences
+    size_t len = str.length();
+    size_t i = len;
+    
+    // Find the start of the last UTF-8 character
+    while (i > 0 && i > len - 4) {
+        i--;
+        unsigned char c = static_cast<unsigned char>(str[i]);
+        
+        // If this is a single-byte character (0xxxxxxx), we're done
+        if ((c & 0x80) == 0) {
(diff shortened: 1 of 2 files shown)


sudhanshu112233shukla:
@rshemet @HenryNdubuaku do check this one as well !!
thank you!!

PR #310, opened 2026-01-31
fix s2t-streaming algorithm

jakmro:


HenryNdubuaku:
Thanks for this @jakmro, Whisper works really well and fast for `cactus transcribe`.  To-Dos:

1. Inspect why moonshine is not working still. 
2. Why NPU doesn't seem to work with `cactus test --ios`

PR #316, opened 2026-02-03
fix: LFM2 multiple tool calls

mhayes853:
Parsing logic was incorrect when lfm2 would emit multiple tool calls.

eg.
```
[{"name":"send_message","arguments":{"recipient":"Blob","message":"Checking the weather in San Francisco.","get_weather(location":"San Francisco"}}]
```

The call for `get_weather` is embedded weirdly in the call for `send_message`.

This PR corrects the parsing logic inside the FFI utils, and adds a new test for handling multiple tool call invocations.

### DIFF
--- a/cactus/ffi/cactus_utils.h
+++ b/cactus/ffi/cactus_utils.h
@@ -420,6 +420,78 @@ inline std::string format_tools_for_prompt(const std::vector<ToolFunction>& tool
     return formatted_tools_json;
 }
 
+static inline std::string trim_lfm2_slice(const std::string& value, size_t begin, size_t end) {
+    while (begin < end && std::isspace(static_cast<unsigned char>(value[begin]))) {
+        begin++;
+    }
+    while (end > begin && std::isspace(static_cast<unsigned char>(value[end - 1]))) {
+        end--;
+    }
+    return value.substr(begin, end - begin);
+}
+
+static inline void append_lfm2_call(const std::string& entry,
+                                   std::vector<std::string>& function_calls) {
+    if (entry.empty()) return;
+
+    std::string trimmed_entry = trim_lfm2_slice(entry, 0, entry.size());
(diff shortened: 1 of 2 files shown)


HenryNdubuaku:
@mhayes853 Thanks for this! does the tool call pass on your end?

```bash
╔══════════════════════════════════════════╗
║          MULTIPLE TOOLS TEST             ║
╚══════════════════════════════════════════╝
├─ User prompt: Send a message to Blob and get the weather for San Francisco.
Response: I'm sorry, I don't have access to a tool to send messages or retrieve weather data directly. However, if you'd like the current weather for San Francisco, I can help with that using the available tool. Let me know how I can assist!

[Results]
├─ Function call: YES
├─ Correct tool: NO
  "success": true,
  "error": null,
  "cloud_handoff": false,
  "response": "I'm sorry, I don't have access to a tool to send messages or retrieve weather data directly. However, if you'd like the current weather for San Francisco, I can help with that using the available tool. Let me know how I can assist!",
  "function_calls": [],
  "confidence": 0.9716,
  "time_to_first_token_ms": 829.07,
  "total_time_ms": 1798.55,
  "prefill_tps": 331.70,
  "decode_tps": 52.61,
  "ram_usage_mb": 51.33,
  "prefill_tokens": 275,
  "decode_tokens": 52,
  "total_tokens": 327
└─ Status: FAILED ✗
✗ FAIL │ tool_multiple_tool_call_invocations
```

mhayes853:
Seems to be model dependent (I've had a 100% success rate with larger models >=1.2B params running on M1 pro). 

If you want, I can loosen the assertion a bit in the test for now to not expect both tool calls outright. I mainly needed it to verify that the parsing logic was fine if the model invoked both tools.

HenryNdubuaku:
Ok makes sense, it seems the 1.2B version is capable of that.

PR #318, opened 2026-02-03
Add Raspberry Pi 5 benchmark (LFM-450M)

michboy:
### Benchmark Results for Raspberry Pi 5

I've added the benchmark results for the **LFM-VL-450M** model.

- **LFM-VL-450M:** 17 tps (Success)
- **LFM-350M:** Failed (Device shutdown/rebooted during test due to power/thermal limits)
- **Moonshine:** Failed (Download error: `pytorch_model.bin` not found)

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ graph.hard_reset();
 | Galaxy A56 | - | - | - |
 | Pixel 6a | 218/44 tps | 2.5s & 36 tps | 1.5s & 189 tps |
 | Nothing CMF | - | - | - |
-| Raspberry Pi 5 | - | 17 tps | - |
+| Raspberry Pi 5 | - | 0.23 & 17 tps | - |
 
  ## Supported Models                                                                                                                                                     
                                                                                                                                                                           


CactusCompute:
@kar-m

kar-m:
What version of Raspberry Pi 5 are you using, how much memory does it have? Also can you paste the full moonshine error and the full test script output for the vl-450m?
@michboy

michboy:
Hi @kar-m,

I tried to re-run the test to capture the full logs, but unfortunately, my Raspberry Pi 5 kept shutting down during the process (likely due to power/thermal limits under sustained load).

However, here is the information I gathered from my initial successful run:

**1. Hardware Specs:**
- **Device:** Raspberry Pi 5
- **Memory:** 8GB RAM
- **OS:** Raspberry Pi OS (64-bit, Bookworm)

**2. LFM2-VL-450M Result:**
- On the first cold run, I observed a stable **17 tok/s**.
- Subsequent runs caused the device to reboot, indicating this model pushes the Pi 5 to its absolute limit even with 8GB RAM.

**3. Moonshine Error:**
- The error was explicitly: `Error: usefulsensors/moonshine does not appear to have a file named pytorch_model.bin`.
- It seems the auto-downloader expects a PyTorch bin file, but the repo structure might be different.

Hope this helps!

kar-m:
Thanks for the details @michboy 

Are you using the official raspberry pi power brick? For my raspberry pi everything else would work really bad, but with the official power brick it was fine running those models, same amount of RAM. However anything above the 450M size will crash the raspberry pi when converting models, so you can try converting a model to cactus weights on a laptop (this will work with a non-ARM laptop) and then sending those over to the raspberry pi, the models should run at least up to 1.6B size IIRC

Where did you see that signature for moonshine? If you saw it in the codebase please add to your PR an edit changing it to `UsefulSensors/moonshine-base` since that is the name they use on HF.

HenryNdubuaku:
nothing has been done about this for a week now @michboy, so I'm closing, but thanks!

PR #319, opened 2026-02-03
Kernel_attention optimization 

Ayan9074:
## Description
Adds a specialized FP16 attention kernel for head_dim == 64, inspired by Qualcomm Hexagon / HVX optimization principles.
Instead of relying on the generic attention path (which handles arbitrary head sizes, masks, and windowing), this change introduces a fixed-shape h64 fast path that:

- Uses compile-time constants for head dimension (64)
- Uses statically allocated accumulators (no dynamic vectors)
- Unrolls the inner dot-product loop
- Applies block-wise numerically stable softmax
- Avoids mask / window logic on decode path

Performance impact tested (M4 MacBook Air) (Average across 10 runs):
Attention 1024×16×64:
Before: ~19.7 ms  (~109 GFLOPS)
After:  ~13.0 ms  (~165 GFLOPS)

Limitations:

- The h64 kernel does not support masks or sliding window attention - falls back to normal version 
- This path is intended specifically where:
    - head_dim == 64
    - No attention mask is present
    - No windowed attention is used

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [X] Performance improvement
- [ ] Documentation update

## Testing
- [X] Tests pass locally
- [X] Tested on ARM hardware
- [X] Benchmarked performance impact

## Checklist
- [X] All commits are signed-off (DCO)
- [X] Code follows project style
- [X] Comments added where necessary
- [ ] Documentation updated if needed

I'm not sure if this is what was being looked for with issue #300 and I haven't updated any documentation(not sure if needed). Is it needed to change this code to adapt for mask and window_size? Should I add this change to cactus_attention_hybrid_int8_fp16 also? Should I just be changing cactus_attention_f16 and not create an extra func for h64?

### DIFF
--- a/cactus/kernel/kernel_attention.cpp
+++ b/cactus/kernel/kernel_attention.cpp
@@ -7,6 +7,143 @@
 #include <cstring>
 #include <vector>
 
+static inline void cactus_attention_f16_h64(
+    const __fp16* queries,
+    const __fp16* keys,
+    const __fp16* values,
+    __fp16* output,
+    size_t batch_size,
+    size_t seq_len,
+    size_t kv_seq_len,
+    size_t num_q_heads,
+    size_t num_kv_heads,
+    float scale,
+    size_t position_offset,
+    bool is_causal
+) {
+    constexpr size_t HEAD_DIM = 64;
+    constexpr size_t BLOCK_SIZE = 32;
+    constexpr float NEG_INF = -INFINITY;
+
+    const size_t group_size = num_q_heads / num_kv_heads;
+    const size_t q_batch_stride = seq_len * num_q_heads * HEAD_DIM;
+    const size_t kv_batch_stride = kv_seq_len * num_kv_heads * HEAD_DIM;
+    const size_t o_batch_stride = q_batch_stride;
+    const size_t q_seq_stride = num_q_heads * HEAD_DIM;
+    const size_t kv_seq_stride = num_kv_heads * HEAD_DIM;
+    const size_t o_seq_stride = q_seq_stride;
+
+    CactusThreading::parallel_for(batch_size * num_q_heads * seq_len, CactusThreading::Thresholds::ATTENTION,
+        [&](size_t start, size_t end) {
+
+        float block_scores[BLOCK_SIZE];
+
+        for (size_t work = start; work < end; ++work) {
+            const size_t batch = work / (num_q_heads * seq_len);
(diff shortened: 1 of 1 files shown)


HenryNdubuaku:
Your head in the right space, does any model benefit from this? Read their attention kernels to see. Also, have you had a look at the HVX issue?

Ayan9074:
yeah some models benefit: FP16 ones like whisper/moonshine/nomic/siglip2, and prefill for gemma/qwen/lfm2 too. but decode for gemma/qwen/lfm2 uses int8 hybrid so this new path won't be used unless I add something for cactus_attention_hybrid_int8_fp16. Also almost all models head_dim is set to 64 so they should be used, and I believe the only model that explicitly passes a window size greater than 0 is gemma but most models still have sliding KV cache by default. Not sure what HVX issue refers to?

HenryNdubuaku:
great @Ayan9074 do you have benchmark numbers before and after?

Ayan9074:
All tests on macbook air m4:

In terms of the perfomance benchmarks on tests:
Attention 1024×16×64:
Before: ~19.7 ms (~109 GFLOPS)
After: ~13.0 ms (~165 GFLOPS)

But since it only really hits audio models, the main impact right now is just for example decode tps rising on my laptop for test like PCM buffer transcription and transcription by around 10% and the total time in ms for the tests to run falls by around 100ms on average. For the longer stream transcription test the total time falls from around 7400ms to 6700ms.

I will work on adapting this so it also affects chat models.

HenryNdubuaku:
@Ayan9074 remember to update me once this is ready

Ayan9074:
@HenryNdubuaku Improvements do occur for Whisper and other FP16 audio models (numbers above), so this PR is ready I believe.

For text models, they use the hybrid INT8 path; I tested an h64 hybrid fast path as well, but decode is dominated by INT8 matmul / memory bandwidth, so it didn’t show meaningful gains. I will keep working on optimizing the attention kernel for the next week to see if there are any other gains to be made, either through SME2 or AMX or HVX style implementations.




Tested on m4 MacBook Air:

Attention 1024×16×64:
Before: ~19.7 ms (~109 GFLOPS)
After: ~13.0 ms (~165 GFLOPS)

Decode tps rising on my laptop for test like PCM buffer transcription and transcription by around 10% and the total time in ms for the tests to run falls by around 100ms on average. 
For the longer stream transcription test the total time falls from around 7400ms to 6600ms.

Issue #295, opened 2026-01-28
Test Flutter (difficulty level: medium)


CactusCompute:
1. Build a simple app that follows the setup instruction [here](https://github.com/cactus-compute/cactus/blob/main/flutter/README.md)
2. Report any bugs and PR documentation fix if poorly written

NyXkim5:
Going to crank this out after midterms tmr! :)

PR #325, opened 2026-02-06
fix: C++20 construct_at compatibility + Flutter bindings bug fixes + test app (#295)

NyXkim5:
## Problem

Building on macOS 14.2.1 with Xcode 15.2 (Apple Clang, `-std=gnu++20`) fails with:

```
error: no matching constructor for initialization of 'cactus::engine::index::Document'
note: in instantiation of ... 'std::__construct_at'
```

The same failure occurs on both macOS and iOS SDK targets. The root cause is that `Document` and `QueryResult` in `engine.h` are plain aggregates, but `emplace_back`/`emplace` calls in `engine_index.cpp` and `cactus_index.cpp` go through `std::construct_at`, which requires an explicit constructor in Apple's libc++ C++20 implementation.

## Fix

Added explicit constructors (with `std::move` for efficiency) and preserved `= default` default constructors for backward compatibility:

- `Document(int, std::vector<float>, std::string, std::string)`
- `QueryResult(int, float)`

## Affected call sites

| File | Line(s) | Call |
|------|---------|------|
| `engine_index.cpp` | 295 | `results.emplace_back(doc_id, embedding, content, metadata)` |
| `engine_index.cpp` | 361, 364 | `top_results.emplace(entry.doc_id, score)` |
| `cactus_index.cpp` | 136 | `docs.emplace_back(...)` |

## Environment

- macOS 14.2.1, Xcode 15.2, Apple Clang with MacOSX14.2.sdk
- Both macOS (`-DCMAKE_SYSTEM_NAME=Darwin`) and iOS builds affected
- Build succeeds with zero new warnings after fix

Refs #295

---
Happy to discuss or iterate on the approach. — Jay Kim


NyXkim5:
## Update: Flutter bindings tested + 5 bug fixes (#295)

Expanded this PR to also cover issue #295 (Test Flutter). Found and fixed 5 critical bugs in `flutter/cactus.dart` that prevented any inference from completing:

1. **Missing Utf8 class** — never defined, compiler fails on every typedef
2. **Non-const default param** — `malloc` as default value is invalid Dart
3. **cactus_init missing 3rd param** — `cache_index` bool dropped from FFI typedef
4. **JSON field name mismatch** — Dart parser expected `text`, C returns `response`; `total_time` vs `total_time_ms`, etc.
5. **Wrong return check** — `!= 0` should be `< 0` (success returns token count)

Added `examples/flutter_test_app` — macOS test app with model loading, completion, and perf metrics.

**Test results (LFM2-350M, macOS Apple Silicon):**
- 99.4 tok/s decode, 453.9 tok/s prefill, 37ms TTFT
- All bugs are blockers — without fixes, zero completions succeed

Closes #295

— Jay Kim

HenryNdubuaku:
hey thanks, but you commited too much files

NyXkim5:
Cleaned up — removed the pre-built cactus.framework binary (2MB) and stripped auto-generated template files (.metadata, widget_test.dart, RunnerTests.swift, pubspec.lock). Also added .gitignore entries to prevent them from coming back.

PR #334, opened 2026-02-08
Add INT8 weight support to conv1d and conv1d_k7s3

KayaanT:
## Summary

Add INT8 weight support to all conv1d variants, with shared dequantization logic and validation.

The generic `conv1d` and specialized `conv1d_k7s3` were previously FP16-only, causing runtime errors for INT8-quantized models (e.g. Whisper, Mamba). The other variants (`conv1d_k3`, `conv1d_causal`) already had INT8 support but with duplicated inline dequantization code.

## Changes

- **Add INT8 weight support to `conv1d` and `conv1d_k7s3`**, handling both grouped and ungrouped quantization with their respective weight layouts (`[C_out, C_in, K]` and `[C_in, K, C_out]`)
- **Extract shared `dequantize_int8_to_fp16` helper** to deduplicate the INT8→FP16 conversion logic across all 4 conv1d variants (`conv1d`, `conv1d_k3`, `conv1d_causal`, `conv1d_k7s3`), reducing ~96 lines of duplicated code
- **Use `W.num_groups` from tensor metadata** instead of recomputing via `K_total / group_size`, with validation that it matches the expected value using ceil division
- **Add grouped INT8 validation**: check that `scales_data` is non-null, `group_size` is non-zero, and `num_groups` is consistent with weight dimensions
- **Add 3 regression tests** comparing INT8 weight paths against FP16 baselines:
  - `conv1d` with ungrouped INT8 weights
  - `conv1d` with grouped INT8 weights and per-group scales
  - `conv1d_k7s3` with ungrouped INT8 weights

## Test plan

- [x] All 41 graph tests pass (including the 3 new ones)


HenryNdubuaku:
@KayaanT thabks for this, what are the implications of this kernel? what does it improve? what are the benchmarks? Did you test quality impact on model?

KayaanT:
> @KayaanT thabks for this, what are the implications of this kernel? what does it improve? what are the benchmarks? Did you test quality impact on model?

looked into this more and the converter doesn't actually quantize 3D tensors so conv1d weights are always FP16. the idea was memory savings (storing as INT8 vs FP16) but benchmarks show no perf or memory gain either since it just dequants to FP16. 

think this can be closed for now and revistied if/when we add native INT8 conv1d support.

justinl66:
Just letting you know I'll be reviewing this @KayaanT ! I'll update soon.

HenryNdubuaku:
@KayaanT I will close this one and create more pressing issues, will update everyone on slack once done!

Issue #320, opened 2026-02-04
Gemma Attention


HenryNdubuaku:
Most models work well with Sliding Window attention only, but Gemma uses a mix of global attention & sliding window attention. Cactus forcefully does not allow window size above 1024, which breaks Gemma beyond that context size as it needs global attention to produce coherent text irrespective of context size. Find a work around.

jrajala6:
Looking into this.

HenryNdubuaku:
Handled by @jrajala6

PR #338, opened 2026-02-10
Fix hybrid attention sliding window for Gemma (#320)

jrajala6:
The hybrid INT8 attention kernel (cactus_attention_hybrid_int8_fp16) wasn't using window_size during decode, so local attention layers in Gemma attended to the full context instead of just recent tokens, breaking the hybrid attention mechanism.

This pr threads window_size from model_gemma.cpp through the graph API down to the kernel. In the kernel, positions outside the window get masked to -inf before softmax. The parameter defaults to 0 (global attention) so nothing changes for other models.

Tested with test_kernel (10/10), test_graph (38/38), and end-to-end chat with Gemma 3 1B.


HenryNdubuaku:
thanks a million for this @jrajala6

PR #340, opened 2026-02-11
Accelerate FP16 matmul via cblas_sgemm for Apple AMX

KayaanT:
## Summary
- Routes large FP16 matmuls (K>=256, M>=4) through Apple Accelerate (`cblas_sgemm`), which uses AMX internally
- Converts FP16 to FP32, calls `cblas_sgemm`, converts result back to FP16
- Small matrices fall through to existing NEON path unchanged
- Same pattern used by the existing conv1d kernel (`kernel_conv.cpp`)
- Reference: [cblas_sgemm](https://developer.apple.com/documentation/accelerate/1513264-cblas_sgemm)

Contributes to #298

## Benchmark (M4 Pro, 1024x1024x1024)

| Benchmark | Before (NEON) | After (AMX) | Speedup |
|---|---|---|---|
| MatMul F16 1024³ | 9.997ms / 215 GFLOPS | 2.524ms / 851 GFLOPS | **4.0x** |
| MatMul 1024³ CPU (graph) | 10.438ms / 206 GFLOPS | 2.419ms / 888 GFLOPS | **4.3x** |
| MatMul F16 1x1024x1024 (GEMV) | 0.046ms / 46 GFLOPS | 0.040ms / 52 GFLOPS | same (NEON path) |

### DIFF
--- a/cactus/kernel/kernel_matmul.cpp
+++ b/cactus/kernel/kernel_matmul.cpp
@@ -3,6 +3,13 @@
 #include <arm_neon.h>
 #include <cstring>
 #include <algorithm>
+#include <vector>
+
+#ifdef __APPLE__
+#include <Accelerate/Accelerate.h>
+constexpr size_t ACCELERATE_M_THRESHOLD = 4;
+constexpr size_t ACCELERATE_K_THRESHOLD = 256;
+#endif
 
 // Do NOT Remove: Uncomment for testing on various paths
 // -----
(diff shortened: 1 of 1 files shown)


HenryNdubuaku:
Thanks for this @KayaanT youve now learnt how we love code contributions, most tasks are designed no need no more than 1-3 file changes, so I'll merge this.

PR #344, opened 2026-02-13
Fix/f16 reduction accum

vyomshah05:
# Fix: FP16 Reduction Accumulation Precision

## Summary

Fixes a numerical precision issue in `cactus_sum_all_f16` where accumulation was performed in FP16, causing potential overflow and silent precision loss for large tensors.

Accumulation is now performed in FP32 while keeping the API unchanged.

---

## Problem

Previous implementation accumulated directly in FP16:

- FP16 max value: 65504
- Limited mantissa precision (~3 decimal digits)

This could cause:
- Overflow for moderate tensor sizes
- Small values being dropped during accumulation
- Silent incorrect results

---

## Fix

Upcast to FP32 immediately after loading and accumulate in FP32:

- Input type remains `__fp16`
- Output remains `double`
- Only internal accumulator precision changed

This matches standard ML framework behavior (PyTorch, cuBLAS, etc.).

---

## Validation

Added stress test summing 32k FP16 ones:

- Previously produced incorrect results
- Now returns the correct sum

---

## Impact

- No API changes
- Improved numerical stability
- Negligible performance impact

### DIFF
--- a/cactus/kernel/kernel_reduce.cpp
+++ b/cactus/kernel/kernel_reduce.cpp
@@ -13,7 +13,6 @@ double cactus_sum_all_f16(const __fp16* data, size_t num_elements) {
             constexpr size_t SIMD_WIDTH = 8;
             const size_t vectorized_end = start_idx + ((end_idx - start_idx) / SIMD_WIDTH) * SIMD_WIDTH;
 
-            // Accumulate in f32 to avoid f16 overflow and precision loss
             float32x4_t sum_lo = vdupq_n_f32(0.0f);
             float32x4_t sum_hi = vdupq_n_f32(0.0f);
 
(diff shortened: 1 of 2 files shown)


ParkiratS:
I looked over this PR and it looks good. Have you found any other instances of this mistake in other parts of the kernel?

vyomshah05:
Yes I noticed the error across the axis and mean functions and made similar fixes and validation tests across them as well

ParkiratS:
Taking a further look into this. Will provide feedback by EOD tomorrow at worst.

ParkiratS:
So there are a few changes we need to make. Firstly, please remove the tests because we do not need the test for this instance since we understand that increased precision will result in higher accuracy. Secondly remove all comments from the code.

vyomshah05:
@ParkiratS I cleaned up the code and tests as you said

ParkiratS:
@vyomshah05 Looks good, I will let Henry know that this is ready to merge.

PR #347, opened 2026-02-13
added only flag

nshejwalkar:
Added an --only flag for testing. Only executes the specified test instead of all tests.


jakmro:
Hi @nshejwalkar, thanks for this PR, looks good. One thing which will improve the maintenance of this flag would be autodetecting the tests inside the test directory.

jakmro:
Thanks for this PR @nshejwalkar!

Issue #299, opened 2026-01-28
ARM SME2 (difficulty level: hard)


CactusCompute:
We need to leverage ARM SME2 for all modern devices, this needs an understanding of Cactus Kernels

Step 1: Review [Cactus Kernels](https://github.com/cactus-compute/cactus/tree/main/cactus/kernel) rigorously

Step 2: Read up ARM SME2 [here](https://www.xda-developers.com/arm-answer-to-npu-something-even-better/) and [here](https://learn.arm.com/learning-paths/cross-platform/multiplying-matrices-with-sme2/5-outer-product/)

Step 3: Figure out kernels in Cactus that could benefit form SME2.

Step 4: Fix one kernel per PR, test with `cactus test` before and after change to verify performance gains. 

If stuck, reach out to @HenryNdubuaku

aarav18:
Working on this.

cattermelon1234:
looking into this as well!

PR #349, opened 2026-02-13
Accelerate MatMul FP16 for ARM SME2

aarav18:
## Summary

- Implements FP16 MatMul using ARM SME2 instructions
- SME2 kernel is only selected if:
  - The compiler supports SME2
  - The CPU supports SME2 (runtime check)
  - `M >= 4` (minimum SME2 tile size)

- Adds a separate compilation target for SME2
  - SME2 requires architecture-specific flags (e.g. `-march=armv9.2-a+sme2`), different from the normal build
  - To avoid double-compilation and impacting global build flags, all SME2 code is isolated to `kernel_sme2.cpp`
  - For future kernels, an "SME2 version" of each kernel file can be created, and all SME2 files can be added to the compilation target
  
- Introduces CMake option `ENABLE_SME2`
  - `AUTO` (default): build SME2 only if supported by the compiler, else build normally
  - `OFF`: Force-disable SME2, build normally

- Existing Neon implementation remains unchanged and serves as the fallback path

- __Note__: I am continuing to work on further optimizations (below) and will add new PRs as efficiency gains are found
  - There is MUCH room for efficiency gains, at least ~2x on top of current speedup

Contributes to #299 

---

## Benchmark (M4)

| Benchmark                          | Before (NEON)           | After (SME2)            | Speedup |
|------------------------------------|--------------------------|------------------------|---------|
| MatMul 1024³ CPU                  | 11.424 ms / 188 GFLOPS   | 2.170 ms / 990 GFLOPS  | 5.3x    |
| MatMul F16 1x1024x1024     | 0.041 ms / 51 GFLOPS    | 0.042 ms / 50 GFLOPS   | ~1.0x (uses NEON) |
| MatMul F16 1024x1024x1024          | 10.819 ms / 199 GFLOPS  | 1.772 ms / 1212 GFLOPS  | 6.1x    |

## Further Optimizations
- Use of multi-vector load/store instructions
- Range prefetching
- Optimizing preprocessing/transposition to leverage SME2 instructions
  - The current `matmul_f16` API uses `(a, b_transposed)` inputs
  - SME2 naturally prefers `(a_transposed, b)` layout, which introduces additional preprocessing overhead

### DIFF
--- a/cactus/kernel/kernel_matmul.cpp
+++ b/cactus/kernel/kernel_matmul.cpp
@@ -179,16 +179,6 @@ void cactus_matmul_f16(
     size_t N
 ) {
 
-#if defined(CACTUS_COMPILE_SME2)
-	if (cpu_has_sme2() && M >= SME2_M_THRESHOLD) {
-		cactus_matmul_f16_sme2_caller(
-			a, b_transposed, c,
-			M, K, N
-		);
-		return;
-	}
-#endif
-
 #ifdef __APPLE__
     if (K >= ACCELERATE_K_THRESHOLD && M >= ACCELERATE_M_THRESHOLD) {
         const size_t a_len = M * K;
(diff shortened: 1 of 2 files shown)


aarav18:
Edited benchmark results after adding 2 commits implementing additional optimizations.

HenryNdubuaku:
@aarav18 thanks so much for this! do we have int8 support.

aarav18:
@HenryNdubuaku Not yet. SME2 does have special instructions for quantized data types like int8, but the implementation is much more complex so I decided to start with fp16.

PR #353, opened 2026-02-15
Vad

jakmro:


HenryNdubuaku:
@jakmro I have optimised the VAD pipeline now, should be 3x faster, study my PR to your branch carefully, to understand the tricks we use to optimise. 

- Fused STFT+magnitude kernel 
- Avoid repeated memory allocation, Pre-allocated process_chunk buffers  & reuse in get_speech_timestamps
- Not every convolution/matmul op benefits from threading, so best to use a threshold to decide when to thread
- Bulk WAV read 
- Windowed-sinc resampler

PR #372, opened 2026-02-20
Improve Telemetry

mhayes853:
Right now, recording telemetry events blocks the current thread in the FFI to make network calls. This can be unacceptable for concurrency runtimes in languages like in Swift Concurrency where blocking for longer than necessary can potentially starve threads in the cooperative thread pool.

Additionally, applications currently don't have an API to explicitly flush or disable telemetry at this time. The former is useful for uploading when an app backgrounds, and the later is useful if apps want to provide a user-specific setting to opt out of telemetry.

This PR does the following:
- Introduces a dedicated worker thread for processing telemetry events, this means that any `record` method will now return immediately.
- Adds `cactus_telemetry_flush` and `cactus_telemetry_shutdown` FFIs to give applications control over the telemetry lifecycle. (`cactus_telemetry_flush` will block until all enqueued events are processed.)
- Adds a new telemetry test suite including an optional integration test that runs when the cloud api key is detected in the environment and `--enable-telemetry` is passed to `cactus test`.
- Uses a global telemetry mutex instead of many atomics to guard the telemetry state. This should prevent higher-level data races from producing inconsistent states.
- Renames `cactus_telemetry_bridge.cpp` -> `cactus_telemetry.cpp` and puts all telemetry related FFIs in there for consistency.

### DIFF
--- a/tests/test_telemetry.cpp
+++ b/tests/test_telemetry.cpp
@@ -89,7 +89,7 @@ bool test_shutdown_then_reinit_then_record() {
 
     cactus::telemetry::shutdown();
     rmdir(cache_dir.c_str());
-    return lines_before_shutdown > 0 && lines_after_reinit > lines_before_shutdown;
+    return lines_after_reinit > lines_before_shutdown;
 }
 
 bool test_record_and_flush_race_no_deadlock() {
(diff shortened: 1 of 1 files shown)


HenryNdubuaku:
I love this PR @mhayes853 it fixed a lot of my concerns with telemetry

PR #374, opened 2026-02-20
added moe support for lfm

kar-m:


rshemet:
super excited about this one

PR #377, opened 2026-02-21
fix: handle spaces in paths when running shell commands

### LINKED ISSUE #379: Bug: cactus build fails when repo path contains spaces
## Description

The `cactus build` command fails when the repository is cloned into a directory path containing spaces (e.g., `/Users/name/Programming Projects/cactus`).

## Error

```
/bin/sh: /Users/name/Programming: No such file or directory
Build failed
```

## Root Cause

The `run_command()` function in `python/src/cli.py` passes string paths directly to `subprocess.run()` with `shell=True`, causing the shell to interpret spaces as argument separators.

## Environment

- macOS (also likely affects Linux/Windows)
- Any path with spaces in the directory name

## Steps to Reproduce

1. Clone the cactus repo into a path with spaces:
   ```bash
   cd "/Users/name/Programming Projects"
   git clone https://github.com/cactus-compute/cactus
   ```
2. Run setup and build:
   ```bash
   cd cactus && source ./setup && cd ..
   cactus build --python
   ```
3. Observe the error

## Suggested Fix

Use `shell=False` with a list argument instead of `shell=True` with a string. This avoids shell interpretation entirely and handles paths with spaces correctly.

adithya-n05:
Fixes #379

## Problem

The `cactus build` command fails when the repo is cloned into a directory path containing spaces (e.g., `/Users/name/Programming Projects/cactus`).

**Error:**
```
/bin/sh: /Users/name/Programming: No such file or directory
Build failed
```

The issue is in `run_command()` which passes string paths to `subprocess.run()` with `shell=True`, causing the shell to interpret spaces as argument separators.

## Solution

Changed `run_command()` to use `shell=False` (the default) by converting string paths to a list. This avoids shell interpretation entirely and handles paths with spaces correctly.

**Before:**
```python
result = subprocess.run(cmd, cwd=cwd, shell=isinstance(cmd, str))
```

**After:**
```python
if isinstance(cmd, str):
    cmd = [cmd]
result = subprocess.run(cmd, cwd=cwd)  # shell=False by default
```

## Why this approach?

- **Cross-platform**: Works on both Unix and Windows (unlike `shlex.quote()` which uses single quotes that break on Windows)
- **Secure**: Avoids shell injection risks by not using `shell=True`
- **Robust**: Handles paths with spaces without any quoting complexity
- **Respects shebangs**: Scripts with `#!/bin/bash` are executed with the correct interpreter

## Testing

Tested with a repo cloned to a path containing spaces:
```
/Users/.../Programming Projects/Hackathons/Cactus AI x Google Deepmind 2026/Resources/...
```

`cactus build --python` now completes successfully.

### DIFF
--- a/python/src/cli.py
+++ b/python/src/cli.py
@@ -5,7 +5,6 @@ import argparse
 import re
 import subprocess
 import shutil
-import shlex
 import platform
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


HenryNdubuaku:
Thanks so much for this @adithya-n05

PR #378, opened 2026-02-21
Optional image cactusrun

### LINKED ISSUE #367: Make Cactus run accept optional image

ishakotalwar:
- Added optional image input support to `cactus run`
- Fixed ANE shape mismatch at the image preprocessing stage 
- updated `README.md`
resolves #367

### DIFF
--- a/cactus/models/model_lfm2vl.cpp
+++ b/cactus/models/model_lfm2vl.cpp
@@ -223,6 +223,7 @@ Lfm2VlModel::MergedEmbeddingResult Lfm2VlModel::merge_image_text_embeddings(
 
     const uint32_t image_start_id = get_token_id("<|image_start|>");
     const uint32_t image_end_id = get_token_id("<|image_end|>");
+    const bool relax_image_placeholder_matching = vision_tower_.use_npu_encoder_;
 
     std::vector<size_t> sequence_nodes;
     sequence_nodes.reserve(tokens.size() + image_embedding_nodes.size());
@@ -283,11 +284,19 @@ Lfm2VlModel::MergedEmbeddingResult Lfm2VlModel::merge_image_text_embeddings(
                 if (inner_token == image_token_id) {
                     flush_segment();
 
-                    if (tile_index < tiles.size()) {
-                        const auto& tile = tiles[tile_index++];
-                        sequence_nodes.push_back(tile.node_id);
-                        total_seq_len += tile.token_count;
+                    if (tile_index >= tiles.size()) {
+                        if (relax_image_placeholder_matching) {
+                            ++token_index;
+                            continue;
+                        }
+                        throw std::runtime_error("More <image> placeholders than projected tile features");
+                    }
 
+                    const auto& tile = tiles[tile_index++];
+                    sequence_nodes.push_back(tile.node_id);
+                    total_seq_len += tile.token_count;
+
+                    if (relax_image_placeholder_matching) {
                         size_t consumed_placeholders = 0;
                         while (consumed_placeholders < tile.token_count &&
                                token_index < tokens.size() &&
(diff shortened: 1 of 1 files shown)


cactus/engine/engine.h
```diff
@@ -802,7 +802,6 @@ namespace index {
             void *mapped_index_, *mapped_data_;
             size_t index_file_size_, data_file_size_;
     };
-} // namespace index
```
kar-m, line 812: revert unnecessary change that doesn't change logic

cactus/engine/engine_image.cpp
```diff
@@ -56,6 +56,12 @@ std::pair<int,int> Siglip2Preprocessor::smart_resize(int height, int width) {
     const int total_factor = config_.patch_size * config_.downsample_factor;
     auto [min_pixels, max_pixels] = compute_pixel_limits();
 
+    if (min_pixels == max_pixels) {
+        int target_side = static_cast<int>(std::sqrt(static_cast<double>(max_pixels)));
+        target_side = std::max(total_factor, round_by_factor(target_side, total_factor));
+        return {target_side, target_side};
+    }
+
```
kar-m, line 64: does this change the logic of the function? the > max_pixels and < min_pixels branches are mutually exclusive already so only one of them gets triggered. What cases are broken without this? Refer to https://github.com/huggingface/transformers/blob/v5.2.0/src/transformers/models/lfm2_vl/image_processing_lfm2_vl_fast.py#L223 smart_resize function for logic that we are porting for the model

cactus/models/model_lfm2vl.cpp
```diff
@@ -318,13 +310,10 @@ Lfm2VlModel::MergedEmbeddingResult Lfm2VlModel::merge_image_text_embeddings(
...
-                    for (size_t remaining = tile_index; remaining < tiles.size(); ++remaining) {
-                        (void)tiles[remaining];
-                    }
-                }
-                throw std::runtime_error("Unused projected tile features remain after processing image block");
+            while (tile_index < tiles.size()) {
+                const auto& tile = tiles[tile_index++];
+                sequence_nodes.push_back(tile.node_id);
+                total_seq_len += tile.token_count;
             }
```
kar-m, line 344: these changes (this whole file) look fine, though i am a little concerned about the removed exceptions. Would they be redundant/wrong?

cactus/models/model_siglip2.cpp
```diff
@@ -50,10 +50,44 @@ void Siglip2VisionModel::load_weights_to_graph(CactusGraph* gb) {
...
+
+                int downsample = std::max(1, static_cast<int>(config_.downsample_factor));
+                int target_image_tokens = npu_input_shape[0] / (downsample * downsample);
+                target_image_tokens = std::max(1, target_image_tokens);
+                preprocessor_config.min_image_tokens = target_image_tokens;
+                preprocessor_config.max_image_tokens = target_image_tokens;
+                preprocessor_config.max_num_patches = npu_input_shape[0];
+
+                preprocessor_ = Siglip2Preprocessor(preprocessor_config);
+            }
```
kar-m, line 90: what are all these new configs for? I don't see them used anywhere else

kar-m: one thing here is that when npu is used we don't actually initialize the siglip model as far as i know, the entire vision encoder runs in the npu and we just populate the kv and conv caches for the lfm2 decoder inside the cactus cpp logic

cactus/models/model_lfm2vl.cpp
ishakotalwar: Those exceptions were correct for before when the counts for placeholder and feature counts were guaranteed to match. But with the ANE fixed-shape runtime preprocessing I think they can become a source of false failure

kar-m: we still want to fully support the non-npu path as well which won't have a guaranteed input shape of 256x256 and the corresponding fixed input shapes

ishakotalwar: that makes sense, i'll fix this

cactus/models/model_siglip2.cpp
ishakotalwar: Does the Siglip2 preprocessor still run before ANE? That's why I added the configs bc they're needed to make the pre-ANE tensor match ANE’s fixed input shape

PR #380, opened 2026-02-21
Justin/api fixes

justinl66:


cactus/ffi/cactus_vad.cpp
```diff
@@ -134,7 +134,7 @@ int cactus_vad(
     try {
         auto start_time = std::chrono::high_resolution_clock::now();
         auto* handle = static_cast<CactusModelHandle*>(model);
-        auto* vad_model = static_cast<SileroVADModel*>(handle->model.get());
+        auto* vad_model = static_cast<SileroVADModel*>(handle->vad_model.get());
```
HenryNdubuaku: This is quite wrong and breaks cactus transcribe.

PR #382, opened 2026-02-21
Add raw FFI binding for Rust

yujonglee:


HenryNdubuaku:
Thanks @yujonglee ! I will tag Hypernote now on the readme as official maintainers if you want.

yujonglee:
@HenryNdubuaku sure
(we named from `hyprnote` to `char` (https://github.com/fastrepl/char))

PR #393, opened 2026-02-23
add --image flag to cactus run for vision models

### LINKED ISSUE #367: Make Cactus run accept optional image

KayaanT:
## Summary
- Adds `--image <path>` optional flag to `cactus run` for vision models
- CLI validates the image path exists before launching the chat binary
- Chat binary accepts optional image path and attaches it to the first user message via the existing `"images"` JSON field

## Usage
```bash
cactus run LiquidAI/LFM2-VL-450M --image photo.png
```

Resolves #367

### DIFF
--- a/python/src/cli.py
+++ b/python/src/cli.py
@@ -787,11 +787,21 @@ def cmd_run(args):
         print_color(RED, f"Error: Chat binary not found at {chat_binary}")
         return 1
 
+    image_path = getattr(args, 'image', None)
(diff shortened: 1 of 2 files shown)


HenryNdubuaku:
@KayaanT thanks so much for this! using an `--image` flag means that the conversation in `cactus run` will only reference an image, but a conversation is supposed to allow referencing multiple images.

PR #395, opened 2026-02-24
add STFT_COMPLEX, derive stft_magnitude via graph composition

yujonglee:
STFT Magnitude: `0.615ms`
STFT Complex: `0.603ms`

in my testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,8 +176,10 @@ git clone https://github.com/cactus-compute/cactus && cd cactus && source ./setu
 - [Android Demo](https://play.google.com/store/apps/details?id=com.rshemetsubuser.myapp)
 
 ## Maintaining Organisations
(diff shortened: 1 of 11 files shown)


HenryNdubuaku:
@yujonglee thanks so much for this! maybe lets swap in the internal logic, of existing fused stft, so we only have one function, to avoid duplicates.

yujonglee:
Hi @HenryNdubuaku  thanks for the review

Done in fe40d278489cd1093a8bb04a018a8ce981f70ce5.

Internal API:
- `cactus_stft_complex_f16` is the single kernel
- `stft_complex` maps directly to it at the graph level
- `stft_magnitude` is now just a graph-level composition
  - `stft_complex` + `slice` + `multiply` + `add` + `sqrt` (no separate kernel)

Public API:
- Unchanged so nothing upstream breaks.

Let me know if I missed something!

Issue #388, opened 2026-02-23
Python bindings: add context manager support for safe resource cleanup


yogyam:
Hey! I noticed that the Python bindings in:

python/src/cactus.py

currently require users to manually call:

cactus_destroy()

to free model memory. This works, but it's pretty easy to accidentally leak memory if an exception gets thrown before the destroy call runs.

For example, looking at:

example.py
```
vlm = cactus_init(str(WEIGHTS_DIR / "lfm2-vl-450m"))
response = cactus_complete(vlm, messages)
embedding = cactus_embed(vlm, "Hello world")
# ... more operations ...
cactus_destroy(vlm)  # never reached if something above throws
```
If cactus_complete or any of the calls between init and destroy raise an exception, cactus_destroy never gets called and the native model memory (which can easily be hundreds of MB or more) leaks with no way to reclaim it. Python’s GC can’t help here since the handle is just an opaque c_void_p.

The standard Python solution for this is context manager support (__enter__ / __exit__). Something like a thin CactusModel wrapper class would make usage both safer and more idiomatic:
```
with CactusModel("weights/lfm2-vl-450m") as model:
    response = model.complete(messages)
    embedding = model.embed("Hello world")
# cactus_destroy called automatically, even if an exception occurred
```
The same issue applies to the index handle (cactus_index_init / cactus_index_destroy) — a CactusIndex wrapper would be useful there too.

This would be fully backward-compatible — the existing free functions would stay as-is, and the wrapper classes would just be an additional, safer way to use the API.

Happy to put together a PR for this if you're interested. Let me know what you think!

Udit8348:
Hey good catch! I think this is worth prototyping as a PR, I am happy to look over it once you have a rough draft. I did a brief look over the code and think your proposed solution seems like a solid initial stating point. Feel free to tag me when you make your PR

Additionally, in your PR can you demonstrate leaky behavior and try to catch it with a tool like `valgrind`. Then, your PR should aim to mitigate this behavior. I believe you can start by looking inside `cactus_ffi.h` to fully catch all allocations. 

Eventually, you can even consider contributing test cases for this.

yogyam:
Hey, just submitted a PR. If you have time I would love for you to look at it. Thanks!
PR #412

PR #412, opened 2026-02-24
feat(python): add context manager support for safe resource cleanup

yogyam:
## What
Based of issue #388 
Added `CactusModel` and `CactusIndex` context manager classes to the Python bindings so `cactus_destroy` / `cactus_index_destroy` are called automatically — even when exceptions occur.

```python
# Before — leaks if anything throws between init and destroy:
model = cactus_init("weights/model")
response = cactus_complete(model, messages)
cactus_destroy(model)

# After — cleanup guaranteed:
with CactusModel("weights/model") as model:
    response = model.complete(messages)
```

Both classes:

* Guard against double-free (calling `destroy()` twice is safe)
* Raise `RuntimeError` on use-after-destroy

---

## FFI Audit

Reviewed `cactus_ffi.h`.

Three handle types allocate native memory:

* `cactus_model_t`
* `cactus_index_t`
* `cactus_stream_transcribe_t`

`cactus_model_t` and `cactus_index_t` receive wrappers.

`cactus_stream_transcribe_t` is a sub-resource of a model (freed by `cactus_stream_transcribe_stop`), so no separate wrapper is needed.

---

## Files Changed

* `python/src/cactus.py`
  Added `CactusModel` and `CactusIndex` classes (appended; no existing code modified)

* `python/src/__init__.py`
  Added lazy imports for new classes

* `python/example.py`
  Updated to use the `with` pattern

* `python/tests/test_context_manager.py`
  Added 8 unit tests (mocked FFI; no model weights required)

---

## Tests

```bash
$ python -m unittest python/tests/test_context_manager.py -v
Ran 8 tests in 0.002s — OK
```

Tests verify:

* `destroy` is called on normal exit
* `destroy` is called on exception
* Double-destroy is safe
* Methods raise after destroy

Manual `leaks --atExit` testing requires a full build + model weights. Happy to add a demo script if needed.

---

## Backward Compatibility

Fully backward-compatible.

All existing free functions remain unchanged. The new classes are purely additive.


Udit8348:
At a glance it looks like its going in a good direction. It does add quite a bit more lines of code, but I can see the intent for it. The backwards compat seems like it should be okay since it does not break the prev syntax style. 

Co-Pilot is asking to add very long and verbose docstrings. I think these are unreadable, imo. I think its better to find a good place to demonstrate how to migrate one's code from the old style to the new one that uses a more concise api. This is also something repo owner (@HenryNdubuaku) should have input on because it is pretty significant.

If you can verify everything with a full build an a model that would be good. Golden, if you can prove your work with some kind of address sanitizer tool. Will depend on your platform. Redundancy checks like this will be helpful for future contribs as well. With that being said, checking for memory leaks using an address sanitizer tool might be trickier than the scope of this issue. Please feel free to investigate.

yogyam:
Thanks for the review. On my recent commits, I added a migration guide to python/README.md showing old vs new pattern. Restored example.py  to its original state — instead of modifying it, I added a separate example_context_manager.py to keep things non-invasive.

I ran the full build verification — cactus build --python + cactus download google/gemma-3-270m-it — and all integration tests pass: normal with-block, cleanup on exception, use-after-destroy protection, and double-destroy safety. The mock-based unit tests also pass. Happy for you or @HenryNdubuaku to run on your end as well.

HenryNdubuaku:
thanks for this @yogyam !
I love this direction! one thing, this is quite chunky and adds a new file to the project, something we generally avoid.

- move the example to the existing example.py
- if we are sticking to context manager, make it the default by refactoring, rather than padding it onto existing flow and bloating the codebase. 
- also update all docs and MD files that use Python.
- CLI depends on the python, review the implications of your PR to these. 
- avoid redundant or unnecessary code at all cost.

Thanks!

yogyam:
Thanks for the feedback! I deleted the extra example file and moved the code into the existing example.py, refactored the Python bindings to make the CactusModel context manager the default API, updated the docs to reflect this, and verified that these changes don't impact the CLI. Let me know what more needs to be done!

HenryNdubuaku:
Thanks for this @yogyam merging now, i decided to retain the old design too, for backward compatibility, while this becomes the new way.

PR #428, opened 2026-02-24
fixing tie capturing for lfm vlm

ncylich:
Renamed ambiguous configs in hf conversion and added checks for both root_config and config to retrieve tie embeddings in accordance with the new Transformers 5.0+ api.

Uses hard links to tie LM head to the embedding weights, eliminating all redundant/wasteful storage use


HenryNdubuaku:
Tests still fail, but I think its because the tree for this branch is dirty now, so I'm closing, lets redo this from a fresh branch @ncylich

PR #436, opened 2026-02-24
feat(stt): WhisperModel logit bias for custom vocabulary

### LINKED ISSUE #396: Add custom vocabulary / hotword biasing for transcription
## Summary

Add support for passing a custom vocabulary (hotword list) to the transcription APIs so that domain-specific words, proper nouns, and jargon are transcribed correctly. Today, both Whisper and Moonshine models run with no way to bias the decoder toward expected words, leading to frequent misspellings of names, acronyms, and technical terms.

## Motivation

On-device transcription is a core Cactus feature (Whisper-Small, Whisper-Medium, Moonshine-Base). In real-world apps — medical, legal, enterprise — users need names like "Cactus Compute", drug names like "Omeprazole", or acronyms like "HIPAA" to be recognized accurately. Without hotword biasing the models hallucinate phonetically similar but incorrect words.

Whisper's `prompt` parameter already provides weak conditioning (the user passes a text prompt that seeds the decoder), but it has limits:
- The prompt is tokenized once and prepended; it cannot scale to hundreds of domain terms.
- It offers no per-token probability boost, so rare words are still suppressed by the language model head.

A dedicated custom-vocabulary mechanism would be more reliable and ergonomic.

### C FFI (`cactus_ffi.h`)

Extend `options_json` accepted by `cactus_transcribe`, `cactus_stream_transcribe_start`, and the SDK wrappers to include a `custom_vocabulary` field:

```jsonc
{
  "max_tokens": 448,
  "custom_vocabulary": ["Cactus", "Ndubuaku", "HIPAA", "Omeprazole"],
  "vocabulary_boost": 5.0   // optional, default ~5.0 logit bias
}
```

No new FFI functions are needed — the vocabulary list flows through the existing `options_json` parameter.

### Engine layer (`engine.h` / `model.h`)

Add a logit-bias hook that fires every decoder step (both `WhisperModel::decode_with_audio` and `MoonshineModel::decode_with_audio`):

1. At init time, tokenize each custom vocabulary entry using the model's `BPETokenizer` / `SPTokenizer` and store the resulting `token_id → boost` map.
2. Before sampling, add the boost value to the logits of every token that appears in any vocabulary entry's tokenization. This is the same pattern already used by `ToolCallConstrainer::get_bias()` and `suppress_tokens_` in `WhisperModel`.

### SDK surface (React Native, Flutter, Kotlin, Swift, Rust, Python)

Each SDK already forwards `options_json` to the C FFI, so they only need:
- Documentation updates showing the new `custom_vocabulary` key.
- Optionally, a typed helper (e.g., `TranscribeOptions.customVocabulary`).

## Implementation Sketch

1. **Parse** `custom_vocabulary` and `vocabulary_boost` from `options_json` inside `cactus_transcribe.cpp` (and `cactus_stream.cpp`), using the same lightweight JSON parsing already in `cactus_utils.h`.
2. **Tokenize** each vocabulary entry via the model's tokenizer. Build `std::unordered_map<uint32_t, float> vocab_bias` mapping each sub-token → boost.
3. **Apply bias** in the decode loop. `WhisperModel` already applies `suppress_tokens_` (sets logits to `-INFINITY`). The custom vocabulary bias is the positive mirror: add `+boost` to the logit of each biased token before temperature scaling / softmax. This should be straightforward given the existing `decode_with_audio` signature already returns via logit manipulation.
4. **Stream transcription** (`cactus_stream_transcribe_start` / `_process`) should accept and persist the vocabulary bias for the lifetime of the stream session.
5. **Tests**: extend `tests/test_stt.cpp` with a case that passes custom vocabulary containing a rare word and asserts it appears in the output.

## Scope & Constraints

- **No model changes** — this is purely a decoding-time logit bias.
- **No new FFI functions** — everything goes through the existing `options_json`.
- Boost values should be clamped to a sane range (e.g., 0–20) to avoid degenerate outputs.
- Works for both Whisper and Moonshine model types.
- The existing `prompt` parameter continues to work alongside custom vocabulary (they are complementary).

## References

- `cactus/ffi/cactus_ffi.h:66-77` — `cactus_transcribe` signature
- `cactus/ffi/cactus_transcribe.cpp:248-286` — decoder loop where bias would be applied
- `cactus/models/model.h:405-667` — `WhisperModel` with existing `suppress_tokens_` logit manipulation
- `cactus/models/model.h:670-829` — `MoonshineModel`
- `cactus/engine/engine.h:395-484` — `ToolCallConstrainer` as prior art for token biasing

vyomshah05:
## What
Adds `vocab_bias_` infrastructure to `WhisperModel` in `model.h` and 
implements the bias application inside `decode_with_audio`.

## Why
Part of the custom vocabulary / hotword biasing feature. Closes #396.

When a caller sets a token → boost map via `set_vocab_bias()`, the decode 
loop executes the graph first, converts logits to FP32, applies the 
clamped boost values, then samples on CPU. When no bias is set, the 
original `gb->sample()` fast path is completely unchanged.

## Changes
- `cactus/models/model.h`: added `vocab_bias_` field and `set_vocab_bias()` 
  public method to `WhisperModel`
- `decode_with_audio`: split sample+execute into fast path (no bias) and 
  manual sampling path (with bias), reusing existing FP32 conversion pattern
- `tests/test_stt.cpp`: added three new tests:
  - `vocab_bias_transcription` — verifies no crash and valid output when 
    `custom_vocabulary` is passed through options_json
  - `vocab_bias_affects_output` — runs same audio with and without extreme 
    bias and logs both outputs for comparison
  - `vocab_bias_direct` — bypasses FFI entirely, calls `set_vocab_bias()` 
    directly on the model to prove the engine path works in isolation

## Not in scope for this PR
- MoonshineModel
- FFI/options_json parsing (needed to connect `custom_vocabulary` JSON → `set_vocab_bias()`)
- SDK docs

## Testing
`cli/cactus test --only stt` passes. No regression on existing transcription.

Note: `vocab_bias_affects_output` currently logs identical outputs because 
the FFI JSON parsing layer is not merged yet — `vocab_bias_` is only 
populated when called directly via `set_vocab_bias()`. The test is 
structured to catch regressions once the FFI wiring lands.

### DIFF
--- a/cactus/engine/engine.h
+++ b/cactus/engine/engine.h
@@ -540,6 +540,10 @@ public:
 
     void* graph_handle_;
 
+    void set_vocab_bias(const std::unordered_map<uint32_t, float>& bias) {
+        vocab_bias_ = bias;
+    }
+
 protected:
     virtual size_t forward(const std::vector<uint32_t>& tokens, bool use_cache = false) = 0;
     
@@ -587,6 +591,9 @@ protected:
     virtual std::vector<__fp16> get_token_embeddings(const std::vector<uint32_t>& tokens);
 
     ToolCallConstrainer tool_constrainer_;
+
+private:
+    std::unordered_map<uint32_t, float> vocab_bias_;
 };
 
 std::unique_ptr<Model> create_model(const std::string& model_folder);
--- a/cactus/engine/engine_model.cpp
+++ b/cactus/engine/engine_model.cpp
@@ -237,7 +237,11 @@ uint32_t Model::decode(const std::vector<uint32_t>& tokens, float temperature, f
     last_hidden = gb->reshape(last_hidden, {1, hidden_dim});
 
     auto logits_node_id = gb->matmul(last_hidden, output_weight_node_id_, true, backend);
-    auto sampled_token_id = gb->sample(logits_node_id, temperature, top_p, top_k, tool_constrainer_.get_bias());
+    auto combined_bias = tool_constrainer_.get_bias();
+    for (const auto& [token_id, boost] : vocab_bias_) {
+        combined_bias[token_id] += boost;
+    }
+    auto sampled_token_id = gb->sample(logits_node_id, temperature, top_p, top_k, combined_bias);
 
     gb->execute(profile_file);
 
--- a/cactus/ffi/cactus_stream.cpp
+++ b/cactus/ffi/cactus_stream.cpp
@@ -470,6 +470,53 @@ cactus_stream_transcribe_t cactus_stream_transcribe_start(cactus_model_t model, 
         CACTUS_LOG_INFO("stream_transcribe_start",
             "Stream transcription initialized for model: " << model_handle->model_name);
 
+        {
+            const std::string opts = options_json ? options_json : "";
+            std::vector<std::string> custom_vocabulary;
+            float vocabulary_boost = 5.0f;
+
+            size_t pos = opts.find("\"vocabulary_boost\"");
+            if (pos != std::string::npos) {
+                pos = opts.find(':', pos);
+                if (pos != std::string::npos) {
+                    ++pos;
+                    while (pos < opts.size() && std::isspace(static_cast<unsigned char>(opts[pos]))) ++pos;
+                    try { vocabulary_boost = std::stof(opts.substr(pos)); } catch (...) {}
+                }
+            }
+
+            pos = opts.find("\"custom_vocabulary\"");
+            if (pos != std::string::npos) {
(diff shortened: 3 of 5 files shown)


HenryNdubuaku:
@vyomshah05 thanks so much, we wanna make sure the solution generalises to all speech models, not just whisper. @ammesatyajit will dive into this as he's leading the efforts at Cactus. The modifications should be here https://github.com/cactus-compute/cactus/blob/38632d3275c46fddc047d7e0dacf42aec9deb789/cactus/engine/engine_model.cpp#L215

Issue #404, opened 2026-02-24
Vectorize variance_axis and eliminate gather in axis reductions


HenryNdubuaku:
## kernel_reduce.cpp — Axis Reductions

| Change | Impact |
|--------|--------|
| **Vectorize `cactus_variance_axis_f16`** — currently fully scalar; use 2-accumulator (sum, sum-of-squares) NEON pass matching the global variance pattern | High — used by LayerNorm and GroupNorm |
| **Eliminate gather pattern in axis min/max/sum** — when `inner_size == 1`, the axis dimension is contiguous and can use straight NEON loads instead of the `values[8]` stack-gather | Medium |

cattermelon1234:
looking into this issue!

PR #460, opened 2026-02-26
Vectorize variance_axis and eliminate gather in axis reductions

cattermelon1234:
# Summary 
- Added fast-path for all axis functions (cactus_min_axis, cactus_max_axis, cactus_mean_axis, cactus_sum_axis) for when inner_size == 1
- Modified cactus_variance_axis to use 2-accumulator NEON pass matching for both inner_size == 1 and other cases (currently all scalar)
- Added 6 benchmark cases to test_performance.cpp, all related to axis reductions (axis sum, mean, max, min, variance), all specifically for benchmarking kernel reductions for inner_size = 1. Also added 1 benchmark for variance when inner_size is not 1
- Added correctness tests specifically for when inner_size == 1 for axis reductions in test_kernel.cpp

Contributes to #404 

# Benchmark (M4 Mac)
## Axis Reduction Performance Comparison (1024x1024)

| Operation                         | Old Time (ms) | New Time (ms) | Speedup (×) | Old GB/s | New GB/s | BW Gain (×) |
|-----------------------------------|---------------|---------------|-------------|----------|----------|-------------|
| Axis Sum                          | 0.067        | 0.038        | **1.76×**   | 31.30    | 55.19    | **1.76×**   |
| Axis Mean                         | 0.055        | 0.038        | **1.45×**   | 38.13    | 55.19    | **1.45×**   |
| Axis Variance                     | 0.327        | 0.058        | **5.64×**   | 6.41     | 36.16    | **5.64×**   |
| Axis Min                          | 0.054        | 0.032        | **1.69×**   | 38.84    | 65.54    | **1.69×**   |
| Axis Max                          | 0.056        | 0.043        | **1.30×**   | 37.45    | 48.77    | **1.30×**   |
| Axis Variance (Non-Inner1 1024x1024x4) | 0.480 | 0.144 | **3.33×** | 17.48 | 58.25 | **3.33×** |

---

### Highlights
- **Variance (inner_size == 1)** improved by **5.6×**
- **Non-Inner1 variance** improved by **3.3×**

# General Comments
- There currently don't exist tests that benchmark variance or tests that benchmark axis reduction where inner size = 1 specifically, so I had to create new tests to benchmark performance
- Some functions in kernel_reduce convert FP16 to FP32 for accumulation, then back to FP16 (i assume for better accuracy), and some do not. Right now, my changes do not convert to FP32, but should I?

### DIFF
--- a/cactus/kernel/kernel_conv.cpp
+++ b/cactus/kernel/kernel_conv.cpp
@@ -34,34 +34,39 @@ static void conv1d_causal_depthwise_f16_accelerate(
         __fp16*       Yb = output + n * out_bs;
         const __fp16* Wc = weight + c * K;
 
-        std::vector<float> padded_input(padded_len, 0.0f);
-        std::vector<float> weight_f32(K);
-        std::vector<float> conv_out(L);
-
-        size_t pad = (K - 1) * dilation;
-        for (size_t t = 0; t < L; ++t) {
-            padded_input[pad + t] = (float)Xb[t * C + c];
-        }
-
-        for (size_t k = 0; k < K; ++k) {
-            weight_f32[k] = (float)Wc[k];
-        }
-
         if (dilation == 1) {
-            vDSP_conv(padded_input.data(), 1, weight_f32.data(), 1,
-                      conv_out.data(), 1, L, K);
-        } else {
-            std::vector<float> dilated_weight(1 + (K - 1) * dilation, 0.0f);
-            for (size_t k = 0; k < K; ++k) {
-                dilated_weight[k * dilation] = weight_f32[k];
-            }
-            size_t dilated_K = dilated_weight.size();
-            vDSP_conv(padded_input.data(), 1, dilated_weight.data(), 1,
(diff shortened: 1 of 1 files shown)


HenryNdubuaku:
thanks @cattermelon1234 for this! did you test? because this does not build, there are serious bugs which I could fix from my end, but then I worry that the code was not properly tested int the first place, should not have worked if you tested. I will close this particular PR to avoid bugs sipping into main. You can submit a fresh tested one, while avoiding AI generated code completely.

cattermelon1234:
> thanks @cattermelon1234 for this! did you test? because this does not build, there are serious bugs which I could fix from my end, but then I worry that the code was not properly tested int the first place, should not have worked if you tested. I will close this particular PR to avoid bugs sipping into main. You can submit a fresh tested one, while avoiding AI generated code completely.

I did test thorougly and the cactus build -> cactus test pipeline worked fine locally. Maybe my local wasn't fully updated with main yet when I tested, I'll do so again locally and reopen when I'm sure it works, sorry about that.

PR #471, opened 2026-02-27
Language detection

ParkiratS:


mshast (commented):
When going through the changes, there were a couple things that caused some errors, so I thought I'd just list them here just in case it causes errors larger than just a local thing.

For some files, I have import errors due to the telemetry header file not being found, but those are an easy fix (adding ../ to those files).  I had a similar issue with is_same_v(), and that might also be a local issue due to compiler versions.

Once those issues were solved, I still had some issues with tests not being able to load. The main test that failed was pcm_transcription because it couldn't find it anywhere, which I think is another local issue, but I'll add to the review if it seems more than just that. With the tests using cactus test, it gave me issues saying it couldn't load VAD despite having it downloaded. Finally, I would see if it's possible to test other languages, as the test in test_stt.cpp is currently only for English.

Issue #490, opened 2026-03-04
cactus convert fails with HF repo id error when merging LoRA adapters


lennartvoelz:
### Summary
Running `cactus convert` on a locally merged LoRA model fails during INT8 conversion with an error about an invalid repo id, because Cactus passes a temporary path to HF where a repo id is expected.

### Console output (abbreviated)
`
Loading base model: google/functiongemma-270m-it
Loading weights: 100%|█████████| 236/236 [...]
Loading LoRA adapter: lora_dir/
Merging LoRA weights into base model...
LoRA merge complete
Saving merged model to temp directory: /var/folders/jl/.../T/cactus_lora_merged_ja9ad9e0
Writing model shards: 100%|█████████| 1/1 [...]

Model weights not found. Downloading /var/folders/jl/.../T/cactus_lora_merged_ja9ad9e0...
=============================================
Converting /var/folders/jl/.../T/cactus_lora_merged_ja9ad9e0 to INT8...
Error: Repo id must be in the form 'repo_name' or 'namespace/repo_name': '/var/folders/jl/.../T/cactus_lora_merged_ja9ad9e0'. Use `repo_type` argument if needed.
Cleaning up temp directory...
`

### Steps to reproduce
1. Install Cactus
2. Prepare a LoRA adapter in lora_dir/
3. Run:
`cactus convert google/functiongemma-270m-it cactus/weights/funcgemma --lora lora_dir/ --precision INT8`

### Suggested fix
Check if the path is a local directory, if so, use the temp directory produced by the merge as the load path

### Remarks
I also found out that the tokeniser model from the LoRA path is not taken by preference

PR #491, opened 2026-03-04
Fix/issue#490

lennartvoelz:
### This merge request fixes issue #490 
The cactus convert pipeline broke in two related ways when working with merged LoRA models:

Tokenizer detection failed: The tokenizer lookup relied on model_id being a Hugging Face repo identifier (e.g. google/functiongemma-270m-it). When merging a LoRA adapter, model_id is set to a local filesystem path, which poisoned name-based detection logic (e.g. special token handling for Gemma models) and caused the sentence piece model lookup to fail entirely.

Merged model treated as HF repo: After merging, the merged weights are written to a temporary directory. The conversion pipeline then attempted to "download" this temp directory as if it were a remote Hugging Face repo ID.

### Fix

**Tokenizer (fix: tokenizer detection failed for LoRA merged models)**
- Introduced a separate name variable for model-name-based detection, decoupled from model_id (which may be a local path)
- The sentence piece model is now copied from the LoRA directory directly, with a fallback to the base model in the HF cache
- Tokenizer config is read locally when available

**LoRA merge path handling (fix: treat merged LoRA temp dir as local model path)**
- Added a check whether model_id is a local directory before invoking the download/conversion step
- If it is a local path (i.e. the merged temp directory), the path is passed directly into weight conversion, skipping the HF download logic entirely

### Impact of the fixes
**Before:**
Bildschirmfoto 2026-03-04 um 15 10 32

**After:**
Bildschirmfoto 2026-03-04 um 15 12 44

-> The output also shows the special gemma tokens now

I ran the test suite and everything works fine

### DIFF
--- a/python/src/cli.py
+++ b/python/src/cli.py
@@ -1603,7 +1603,8 @@ def cmd_convert(args):
         merged_model.save_pretrained(temp_merged_dir)
         tokenizer.save_pretrained(temp_merged_dir)
 
-        # Copy SentencePiece .model file: prefer LoRA adapter's, fall back to base model's
+        # Copy SentencePiece .model file and tokenizer_config.json if they exist
+        # in the LoRA adapter directory
         lora_sp = next(Path(lora_path).glob("*.model"), None)
         if lora_sp:
             shutil.copy2(lora_sp, Path(temp_merged_dir) / lora_sp.name)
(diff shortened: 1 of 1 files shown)


lennartvoelz:
I added one more fix, so that the tokenizer config is also copied from the LoRA directory if present
Bildschirmfoto 2026-03-04 um 15 32 14

HenryNdubuaku:
thanks for this @lennartvoelz !!!

PR #493, opened 2026-03-04
Add Gemma 3N (text-only) model support

ncylich:
## Summary
- Adds text-only Gemma 3N (E4B) inference with AltUp (4-stream alternating updates), Laurel (learned augmented residual layers), and Per-Layer Input (PLI)
- Implements hybrid local/global attention with sliding window, KV-cache sharing for the last 10 layers, per-head QK normalization, V normalization, and Gaussian top-k activation sparsity
- Adds weight conversion pipeline and config parsing for `gemma3n` model type

### DIFF
--- a/cactus/graph/graph.h
+++ b/cactus/graph/graph.h
@@ -140,7 +140,8 @@ enum class OpType {
     GATED_DELTANET_PREFILL,
     STFT,
     ALTUP_PREDICT,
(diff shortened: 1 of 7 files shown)


HenryNdubuaku:
@ncylich we need to resolve the conflicts, also you need to add gemma 3n to the readme

PR #494, opened 2026-03-05
add Android benchmarking harness with battery and thermal metrics

Vinesh2929:
Extends test coverage to mid-range and budget Android devices that are currently missing from the benchmarks table. Adds battery drain, thermal throttle detection, and OOM event tracking - metrics that matter for always-on edge inference but aren't tracked by any existing mobile benchmark tool.

New files under tests/android_bench/:
- bench.py: main runner - ADB push, 3 warmup + 5 measurement iters, exponential-backoff retry, atomic result writes, SIGINT handler
- adb_client.py: single ADBClient class wrapping all device I/O
- metrics.py: cactus JSON parser + battery/thermal/OOM collectors
- renderer.py: README-format table, extended table, JSON artifact
- scenarios.yaml: LFM 1.2B, LFMVL 1.6B, Parakeet 1.1B, INT8 sweep
- devices.yaml: registry template for 10 target devices
- tests/: 60 unit tests (parser + renderer), no device required
- docs/android_benchmarks.md: how to run + reference numbers

**New metrics nobody currently tracks:**
- `battery_drain_pct` — battery % consumed during 5-min sustained inference load
- `thermal_throttle_detected` — flags >15% CPU clock drop during run (Snapdragon threshold)
- `thermal_peak_celsius` — peak temperature across all thermal zones
- `memory_pressure_events` — OOM/lmkd events isolated to the benchmark window

**Engineering details:**
- 3 warmup + 5 measurement iterations with exponential-backoff retry
- `ThermalPoller` daemon thread samples every 2s without blocking benchmark loop
- Atomic result writes after every scenario — ADB disconnect loses at most one scenario
- SIGINT handler saves progress before exit
- Renderer output matches README table format exactly — rows paste in without editing

### Device results
Currently validating against real devices via Firebase Test Lab — results to 
follow. Happy to prioritize whichever devices are most useful to the team.

### DIFF
--- a/tests/android_bench/adb_client.py
+++ b/tests/android_bench/adb_client.py
@@ -1,0 +1,197 @@
+"""
+ADB wrapper for the Cactus Android benchmarking harness.
+
+All device interactions go through ADBClient so error handling and
+retry logic stay in one place, rather than scattered across the runner.
+"""
+
+import logging
+import subprocess
+import time
+from typing import Optional
+
+logger = logging.getLogger(__name__)
+
+
+class ADBError(Exception):
+    """Raised when an ADB command fails or the device is unreachable."""
+
+
+class ADBClient:
+    """
+    Thin wrapper around the `adb` CLI for a single connected device.
+
+    Every shell command is logged at DEBUG level with the device serial so
+    multi-device runs produce unambiguous logs. Command timeouts are enforced
+    to prevent the benchmark runner from hanging if the device crashes.
+    """
+
+    def __init__(self, serial: str, default_timeout: int = 60) -> None:
+        """
+        Initialise the client and verify the device is reachable.
+
+        Args:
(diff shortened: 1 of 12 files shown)


tests/android_bench/adb_client.py
```diff
@@ -0,0 +1,197 @@
...
+    def is_connected(self) -> bool:
+        """Return True if the device serial still appears in `adb devices`."""
+        try:
+            result = subprocess.run(
+                ["adb", "devices"],
+                capture_output=True,
+                text=True,
+                timeout=5,
+            )
+            return self.serial in result.stdout
```
FarooqMulla: substring match on adb devices output. Serial 1234 false-matches device 12345678. Parse line-by-line and match the first whitespace-delimited token exactly.

PR #502, opened 2026-03-06
fix: make FunctionGemma prompt formatting strict

lennartvoelz:
- Remove hardcoded “When you decide to call…” guidance and arg example from developer turn
- Fix FunctionGemma trigger string (no trailing period) and append tools declarations directly
- Wrap tool responses in a developer turn and allow stacking multiple tool responses
- Stop wrapping tool outputs in value:; pass through {...} payload as-is
- Close pending tool-response developer turn before next user/model turn to avoid malformed prompts

[Follows the official docs](https://ai.google.dev/gemma/docs/functiongemma/formatting-and-best-practices)

### DIFF
--- a/cactus/engine/engine_tokenizer.cpp
+++ b/cactus/engine/engine_tokenizer.cpp
@@ -331,10 +331,8 @@ std::string Tokenizer::format_gemma_style(const std::vector<ChatMessage>& messag
             }
         }
         if (!tools_json.empty()) {
-            result += "You are a model that can do function calling with the following functions.";
+            result += "You are a model that can do function calling with the following functions";
             result += tools_json;
-            result += "\n\nWhen you decide to call a function, output it in this exact format:\n";
-            result += "<start_function_call>call:function_name{arg1:<escape>value1<escape>,arg2:<escape>value2<escape>}<end_function_call>";
         }
         result += "<end_of_turn>\n";
     }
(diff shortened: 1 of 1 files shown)


HenryNdubuaku:
@lennartvoelz thanks for this! one thing, this fails more tool call tests than the old setup, any insights as to why?

lennartvoelz:
What did you test it against? @HenryNdubuaku

lennartvoelz:
@HenryNdubuaku 

I would say for the standard weights, it is just noise. But there are a few things off with FunctionGemma setup rn. I'll later push some changes that make the model much more stable. As said, I tested against the Hackathon dataset. With my changes, the effect is especially visible for fine tuned models.
**Fine-tuned & new setup:**
Bildschirmfoto 2026-03-07 um 14 13 23
**Fine-tuned & old setup:**
Bildschirmfoto 2026-03-07 um 14 13 47
**Base weights & new setup:**
Bildschirmfoto 2026-03-07 um 14 14 27
**Base weights & old setup:**
Bildschirmfoto 2026-03-07 um 14 14 02

With the changes in the Gemma implementation, the effect of the extra tokens on latency is much smaller. However, I believe it is better practice to keep it standard (which also works better in this case!).
I am still working on reducing the execution time while maintaining the much better sampling quality.

HenryNdubuaku:
thanks @lennartvoelz I will merge this now, thanks for testing

Issue #503, opened 2026-03-06
FunctionGemma FP16 issues


lennartvoelz:
### Summary

Gemma3 models (including FunctionGemma-270M) produce incorrect outputs when run in FP16 precision through the Cactus inference engine. The root cause is **FP16 overflow** (values exceeding the 65504 max) at multiple points in the computation graph that destroy model outputs.

This issue is **specific to FP16** and does not occur with BF16 or FP32, because BF16 supports the same exponent range as FP32 (max ~3.4e38) while FP16 is limited to 65504.

References:
- [Gemma 3 is broken with fp16](https://github.com/huggingface/transformers/issues/36822)
- [BF16 vs. FP16 vs. FP32 for Gemma 3 Inference — Mind Your Data Type](https://kaitchup.substack.com/p/mind-your-data-type-bf16-vs-fp16)

### Why Gemma3 architecture is uniquely affected
1. Very large RMSnorm weights -> activations after norm can easily overflow FP16 range
2. Post norm (attention, pre/post ffn) -> multiple possible overflow points in each transformer block
3. Gate projection can overflow before activation functions compress
4. Large vocab

### Analysis Chain
I first checked that the prompt/tool formatting were in agreement with the official FunctionGemma docs (see #501). Then I made sure, that the model conversion and LoRA merging worked as intended and that the correct tokenizer with the special Gemma tokens was used. Next, I scanned the implementation code to ensure that the model was implemented correctly.
When everything was in place (after some fixes), I verified the correctness of the model (with LoRA merged weights) by using the standard Hugging Face transformers library in Python. When loading the exact same model weights and prompt via AutoModelForCausalLM using torch.float16, the model instantly collapses into a <pad> loop. When loaded to torch.float32, the exact same script produces perfectly formatted tool-calling output.

Since Cactus mostly supports FP16 operations, I performed an activation analysis on all sensible operations (really useful debug feature on graph level btw). I observed the following (potential, not for all inputs) overflow locations:
1. Residual stream after layer 5+
2. RMSnorm outputs: overflows when weight values are large
3. Gate and up projections produce values that, when multiplied together after GELU, exceed FP16 range
4. Post-attention / post-MLP norms
5. LM head

### Why this is even worse in fine-tuned models (LoRA)
While base Gemma 3 models occasionally trip the 65k limit, this issue is more severe in fine-tuned models.
[...]

PR #508, opened 2026-03-08
Fix/issue#503: FunctionGemma now usable

lennartvoelz:
This PR fixes issue #503 
After trying overflow prevention in the residual stream, I realized that the issue is the logit sampling. Sampling
directly in FP16 causes numerical breakdown: the softmax and probability normalization lose too much precision across the 256K+ vocabulary, leading to degenerate token selection (repetition, gibberish, wrong/hallucinated tool calls).

I introduced a new sampling kernel, that reads FP16 logins and performs all internal math in fp32 (I previously tried to use the existing cactus_sample_f32 kernel, but the upcasting introduced too much overhead for additional fp32 buffer allocs). This kernel is only used, when a flag (fp32_accumulate in OpParams) is set, so other models that do not need the higher precision in sampling math are not affected.

This is the rough strategy:
- Fused widening convert with scaling and running max acc
- Reduce to a smaller active token list to reduce working set
-> Remaining work on the really small list
- Allocations: all buffers are static thread_local and reused across tokens
- RNG: static thread_local std::mt19937, seeded once per thread

The kernel is a bit slower but the overhead vs the pure f16 kernel is dominated by BW in the full vocab pass, so basically unavoidable (at least I think so, correct me if I'm wrong).

The result is a much better working model with roughly 370ms average latency (MacBook M3) on the Hackathon dataset. Before the changes, fine tuned models were unusable (described in #503). The new kernel also make the base weight model much better.

Here are the results from the "benchmark":
Fine-tuned & new sampling:
Bildschirmfoto 2026-03-08 um 00 54 58
Fine-tuned & old sampling:
Bildschirmfoto 2026-03-08 um 00 58 40
Base weights & new sampling:
Bildschirmfoto 2026-03-08 um 00 56 39
Base weights & old sampling:
Bildschirmfoto 2026-03-08 um 00 58 02

Please note, that this is not a real benchmark in terms of the model latency, because the hackathon benchmark setup returns 0ms on model failure -> the times for the old sampling are heavily skewed towards smaller execution times.
I did not have time to properly benchmark the execution times. If you want me to do that, let me know


lennartvoelz:
With all the latest fixes applied (see #510), the fine-tuned model now matches the outputs from the Transformers library's reference implementation. This allows all Gemma family models to be fine-tuned and deployed in Cactus :)
This was not possible before the chain of fixes.
Benchmark now with the fine tuned model:
Bildschirmfoto 2026-03-08 um 22 59 44
Benchmark now with the base FunctionGemma model:
Bildschirmfoto 2026-03-08 um 22 57 32

Who needs cloud fallback now...

HenryNdubuaku:
@lennartvoelz thanks, you wanna resolve the conflicts

lennartvoelz:
Replaced with #510

PR #560, opened 2026-04-07
Expose min_p and repetition_penalty in completion options

DuFanYin:
Fixes #554

> Scope note: This PR intentionally lands the minimal option-exposure fix first. If preferred by maintainers, I can add follow-up commits in this PR to refactor multimodal sampling paths (and any API-shaping changes) for fully consistent repetition behavior.

**Approach**

Did not change the public C function signatures: new fields live in `options_json` and flow through `InferenceOptions`. Internal C++ (`Model::decode`, `decode_with_images`, `decode_with_audio`, `sample_token`, graph `OpParams`) gains parameters with defaults so existing call sites keep building without changes.

**What changed**

- `InferenceOptions`: add `min_p` (default `0.15`) and `repetition_penalty` (default `1.1`); extend `parse_inference_options_json` to parse both keys, keeping defaults when omitted.
- FFI: pass both through all `decode` / `decode_with_audio` call sites in `cactus_complete` and `cactus_transcribe`, so the expanded `Model::decode*` signatures receive the right arguments from `options_json`.
- Graph: extend `OpParams`; add `CactusGraph::sample_with_options`. Existing `sample()` forwards to it with the prior implicit defaults so unmodified call sites are unaffected.
- Kernel: add `cactus_sample_f32_ex` / `cactus_sample_f16_ex`; keep `cactus_sample_f32` / `cactus_sample_f16` as wrappers for backward compatibility.
- `min_p`: now applied on both FP32 and FP16 paths (FP16 previously skipped this). Standard rule: after temperature / top-k / bias, compute probabilities and drop logits below `max_prob * min_p`.
- `repetition_penalty`: implemented in `Model::sample_token` via the existing logit-bias map — subtract `log(penalty)` for each token in `token_history_`. History is bounded to 128 tokens per instance and cleared on `reset_cache()`.
- Remove `static token_history` from the FP16 sampler (process-wide state, broken for multiple concurrent sessions).
- `docs/cactus_engine.md`: document both options and defaults.

**Design notes for reviewers**

- **`repetition_penalty` is intentionally a no-op at the kernel level.** `cactus_sample_f32_ex` / `cactus_sample_f16_ex` accept the parameter but discard it (`(void)`). The penalty is fully applied upstream in `Model::sample_token` as a logit bias before the graph executes. The kernel parameter exists for API symmetry and future direct-graph use cases. This is not a bug.
- **`sample()` now implicitly activates `min_p=0.15`.** The old `sample()` had a hardcoded `constexpr float min_p = 0.15f` internally. The new `sample()` forwards to `sample_with_options` with the same value, so behavior is unchanged for existing callers.

- **Whisper language probe remains greedy and option-independent.** The initial language probe calls `decode_with_audio` with `temperature=0`, `top_p=0`, `top_k=0`, and explicitly passes `min_p=0` / `repetition_penalty=1` so user decoding settings do not affect language detection.
- Known limitation (pre-existing): some multimodal decode paths sample directly from graph nodes (without Model::sample_token), so repetition behavior is not fully unified with the standard text decode path in this PR.

---

### DIFF
--- a/cactus/engine/engine.h
+++ b/cactus/engine/engine.h
@@ -647,6 +647,12 @@ public:
     virtual std::vector<float> get_audio_embeddings(const std::vector<float>& audio_features);
 
     virtual void reset_cache() { kv_cache_.reset(); token_history_.clear(); }
+    void record_sampled_token(uint32_t token) {
+        if (token_history_.size() >= MAX_TOKEN_HISTORY) {
+            token_history_.erase(token_history_.begin(), token_history_.begin() + (MAX_TOKEN_HISTORY / 2));
+        }
+        token_history_.push_back(token);
+    }
 
     double score_tokens_window_logprob(const std::vector<uint32_t>& tokens, size_t start, size_t end, size_t context, size_t* tokens_scored);
 
--- a/cactus/engine/engine.h
+++ b/cactus/engine/engine.h
@@ -622,7 +622,8 @@ public:
               const std::string& system_prompt = "", bool do_warmup = true);
 
     virtual uint32_t decode(const std::vector<uint32_t>& tokens, float temperature = -1.0f, float top_p = -1.0f,
-                      size_t top_k = 0, const std::string& profile_file = "", float* out_entropy = nullptr);
+                      size_t top_k = 0, const std::string& profile_file = "", float* out_entropy = nullptr,
+                      float min_p = 0.15f, float repetition_penalty = 1.1f);
 
     virtual void prefill(const std::vector<uint32_t>& tokens, size_t chunk_size = 256, const std::string& profile_file = "");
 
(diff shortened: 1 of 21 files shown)


HenryNdubuaku:
@yujonglee wanna review this one? Thanks @DuFanYin !

yujonglee:
I think most of the file touches are justified by the chosen plumbing path, especially for `min_p`.

That said, I don’t think the implementation is fully consistent yet:

- `Lfm2VlModel::decode_with_images()` still samples directly via `gb->sample(...)` on the image path, so it bypasses the new `min_p` / `repetition_penalty` handling that now lives in `Model::sample_token()`. It also looks like the sampled token is not recorded back into history on that path.
- `Gemma4MmModel::decode_multimodal()` now threads `min_p`, but it still samples directly instead of going through `Model::sample_token()`, so `repetition_penalty` is still not honored on that multimodal path.

DuFanYin:
> I think most of the file touches are justified by the chosen plumbing path, especially for `min_p`.
> 
> That said, I don’t think the implementation is fully consistent yet:
> 
> * `Lfm2VlModel::decode_with_images()` still samples directly via `gb->sample(...)` on the image path, so it bypasses the new `min_p` / `repetition_penalty` handling that now lives in `Model::sample_token()`. It also looks like the sampled token is not recorded back into history on that path.
> * `Gemma4MmModel::decode_multimodal()` now threads `min_p`, but it still samples directly instead of going through `Model::sample_token()`, so `repetition_penalty` is still not honored on that multimodal path.

fixed, They now use language_model_.sample_token(...) plus record_sampled_token(...), matching text decode.

PR #562, opened 2026-04-07
fix apple i8mm detection to use runtime sysctl check

DuFanYin:
## Problem
On Apple, `cpu_has_i8mm()` was hardcoded to `has = true`, which could incorrectly enable i8mm kernels on unsupported hardware.  
This is causing illegal-instruction crashes instead of safely falling back to non-i8mm paths.

## History
- Early implementation (old `cactus_has_i8mm()` path, e.g. `a8b3f013`) used hardcoded Apple support.
- Runtime detection was later introduced in `1e3f7b58` using:
  `sysctlbyname("hw.optional.arm.FEAT_I8MM", ...)`.
- During refactor in `c38aec77` (`cpu_has_i8mm()`), Apple logic regressed back to:
  `has = true`.

## Fix
- Restore runtime detection in `cpu_has_i8mm()` for Apple.
- Replace hardcoded `has = true` with `sysctlbyname("hw.optional.arm.FEAT_I8MM", ...)`.
- Keep existing `std::call_once` probe pattern and conservative default (`false` if probe fails).

### DIFF
--- a/apple/CmakeLists.txt
+++ b/apple/CmakeLists.txt
@@ -242,7 +242,7 @@ if(BUILD_SME2)
     target_compile_definitions(cactus PRIVATE CACTUS_COMPILE_SME2)
 endif()
 
-# I8MM support: all Apple Silicon supports I8MM
+# I8MM support: detected at runtime via sysctl, not all Apple Silicon has FEAT_I8MM
 add_library(cactus_i8mm_obj OBJECT ${I8MM_SOURCE})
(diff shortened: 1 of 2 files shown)


DuFanYin:
@HenryNdubuaku, Hi — this fixes a bug where INT8 matmul could run the I8MM path (e.g. vmmlaq_s32) on Apple ARM64 devices that do not report FEAT_I8MM, leading to an illegal-instruction crash. I skipped filing an issue and went straight to a PR.

PR #569, opened 2026-04-10
Follow-up: consolidate sampling APIs after #560

DuFanYin:
## Context
PR #560 (`e9cc4684e389a689002c27d7478f8b2d5024f8b7`) exposed `min_p` and `repetition_penalty` through completion/model paths, including text, VLM, and ASR decode flows.

To stay minimally disruptive, it kept compatibility wrappers and compromise argument ordering at some lower layers.

## What this follow-up changes
- **Unify sampling argument order** to semantic grouping across decode/sample paths:
  - from: `temperature, top_p, top_k, profile_file, out_entropy, min_p, repetition_penalty`
  - to: `temperature, top_p, min_p, repetition_penalty, top_k, profile_file, out_entropy`
  - applied across `decode()`, `decode_with_images()`, `decode_with_audio()`, and `sample_token()` call chains.
  - **rationale**: group all sampling distribution parameters (`temperature`, `top_p`, `min_p`, `repetition_penalty`) together before output-shaping/diagnostic parameters (`top_k`, `profile_file`, `out_entropy`). This makes call sites easier to read and reduces the risk of silently swapping `float` arguments with similar types.
- **Remove legacy wrapper APIs** introduced for transition:
  - `cactus_sample_f32_ex` -> `cactus_sample_f32`
  - `cactus_sample_f16_ex` -> `cactus_sample_f16`
  - `CactusGraph::sample_with_options()` -> `CactusGraph::sample()`
- **Drop dead `repetition_penalty` params from stateless layers** (kernel/graph/C graph FFI), where token history is unavailable and the value was not meaningfully applied.
  - keep `repetition_penalty` where it is actually enforced (`Model::sample_token()` via logit-bias path)
  - keep `min_p` in kernel/graph surfaces (it is executed there)
- **Update docs and C/FFI signatures** to match the consolidated API surface.
- **Fix Gemma4 tests** to match the new decode argument order/signatures (`tests/test_gemma4_audio.cpp`, `tests/test_gemma4_suite.cpp`, `tests/test_gemma4_vision.cpp`).

## Behavior impact
No intended behavior change in generation/transcription outputs; this is API cleanup and consistency work on top of #560.

### DIFF
--- a/cactus/engine/engine.h
+++ b/cactus/engine/engine.h
@@ -627,9 +627,10 @@ public:
     virtual bool init(CactusGraph* external_graph, const std::string& model_folder, size_t context_size,
               const std::string& system_prompt = "", bool do_warmup = true);
 
-    virtual uint32_t decode(const std::vector<uint32_t>& tokens, float temperature = -1.0f, float top_p = -1.0f,
-                      size_t top_k = 0, const std::string& profile_file = "", float* out_entropy = nullptr,
-                      float min_p = 0.15f, float repetition_penalty = 1.1f);
+    virtual uint32_t decode(const std::vector<uint32_t>& tokens,
+                      float temperature = -1.0f, float top_p = -1.0f, float min_p = 0.15f,
+                      float repetition_penalty = 1.1f, size_t top_k = 0,
+                      const std::string& profile_file = "", float* out_entropy = nullptr);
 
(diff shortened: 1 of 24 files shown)


DuFanYin:
@yujonglee Hi, this is a follow up refactor/cleanup with no behavior change, could you help me review this? Thx!

PR #578, opened 2026-04-14
Graph remaining ops

cattermelon1234:
# Summary
refactored graph API, implemented all ops supported by cactus graph into the python API. Before, I had a very large bitmask to represent the different parameters of each OpType, and I had separate write functions for each optype, as well as different serialization functions. After implementing the remaining ops, I realized that it would be extremely difficult to update the Python API if a new cactus graph kernel is ever implemented. Thus, I refactored the code. 

I also implemented wrappers for all of the remaining functions supported by the cactus graph into the python API (ex: attention, conv, sample, permute, etc.) 

The process is still kind of lengthy, but it's more streamlined. There is a set table of all ops and their external parameters, and the user just has to update that table and implement all functions for load/save to work correctly. 

### Contributing Graph Operations
1. **Define the op in core graph types** 
Add the new OpType in `cactus/graph/graph.h`
If the op needs additional parameters, add the fields to OpParams in the same file 

2. **Add a graph builder API**  
Add a builder method in `cactus/graph/graph_builder.cpp` and its declaration in 
`cactus/graph/graph.h`
Follow the pattern of existing builder methods, e.g. for a new "relu" op:
```cpp
size_t CactusGraph::relu(size_t input) {
    OpParams params;
    return add_node(OpType::RELU, {input}, params);
}
```
3. **Implement the op in the execution engine**
Implement the krnel or graph op code in the relevant file, usually in `cactus/kernel/`
Register the new op in the dispatch table in `cactus/graph/graph_execute.cpp` for the supported backends (CPU, NPU)

4. **Export op in FFI bindings**
- header: `cactus/ffi/cactus_ffi.h`
- implementation: `cactus/ffi/cactus_ffi.cpp`

5. **Add python ctypes declaration** 
Add `_lib.cactus_graph_my_new_op.argtypes/restype` in `python/src/cactus.py`

6. **Add python graph wrapper** 
Add `Graph.my_new_op(...)` in `python/src/graph.py`, and optionally a Tensor
  convenience method.

7. **Add serialization schema entry if needed**
If your op has extra parameters that need to be saved/loaded that are not in the 
default node, add new ParamField enum values. 

If the op has any graph-persistent params:

- add any new ParamField enum values in cactus/graph/graph_param_io.cpp
- add read/write logic for those fields
- add the op’s schema entry in `op_schema(...)`

If the op has no params, you may not need to touch schema beyond maybe adding an
empty schema entry.

The syntax pattern there is:
```
{OpType::MY_NEW_OP, {
    {ParamField::Alpha, FieldPersistence::Persistent},
    {ParamField::Mode, FieldPersistence::Persistent},
}},
```
8. **Add test coverage**
Add unit tests to `tests/test_graph.cpp` covering the native graph function, and 
add python tests in `python/tests/test_graph.py` covering the Python API and end-to-end execution.

and then support those fields in `write_field(...)` / `read_field(...)`.

If a field is runtime-only, mark it RuntimeOnly instead of Persistent.

### DIFF
--- a/cactus/graph/graph_io.cpp
+++ b/cactus/graph/graph_io.cpp
@@ -1,4 +1,5 @@
 #include "graph.h"
+#include "graph_param_io.h"
 #include <fstream>
 #include <stdexcept>
 #include <sys/mman.h>
@@ -35,14 +36,6 @@ namespace {
       out.write(reinterpret_cast<const char*>(&v), sizeof(v));
     }
 
-    inline void write_i32(std::ostream& out, int32_t v) {
-      out.write(reinterpret_cast<const char*>(&v), sizeof(v));
-    }
-
-    inline void write_f32(std::ostream& out, float v) {
-      out.write(reinterpret_cast<const char*>(&v), sizeof(v));
-    }
-
     void write_size_vector(std::ostream& out, const std::vector<size_t>& values) {
       uint32_t size = static_cast<uint32_t>(values.size());
       write_u32(out, size);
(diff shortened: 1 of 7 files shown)


cattermelon1234:
not included, but as a full end-to-end verification I recreated model_gemma.cpp with the python/pytorch exposed API, loaded the weights, and it produced the same top token predictions for a prompt, which is a good sign that the pytorch api is a lossless conversion from the cpp api. 

i'm mostly finished with the pytorch -> cactus graph transpiler, but I was missing a lot of key ops so I paused that to work on this PR first. Ideally, the final version of this pipeline is a user can train their own custom model through pytorch, pass the file and weights into cactus, and run inference.

PR #598, opened 2026-04-18
Gemma4 multimodal multi-turn fixes

ncylich:
## Summary

Fixes multi-turn multimodal chat for Gemma 4 so audio asked *after* an image turn actually reaches the model (earlier it kept describing the image instead of answering).

## What changed

- **Audio placeholder restoration.** Cache each user turn's audio soft-token count on the handle; the template re-emits the same placeholder block on subsequent turns without reloading or re-encoding prior audio.
- **`decode_multimodal` cleanup.** Dropped the `prefill_completed_` / `last_token_count_` bookkeeping (which the text/image decode path was silently invalidating). Now uses `kv_cache_.current_seq_len` as the single source of truth and dispatches to `forward_multimodal` only when the delta actually contains image or audio placeholder tokens.
- **BPE whitespace-only merges.** `merges.txt` uses `\n` as both line terminator and merge content, so merges like `\n + \n = \n\n` can't be encoded in that format and were silently dropped. Recover them by scanning the vocabulary after load for pure-whitespace tokens and synthesizing the corresponding merge rules. This restores `encode(decode(tokens)) == tokens` for the common markdown/paragraph cases.
- **chat.cpp image behaviour.** The persistent ``current_image`` was being re-attached to *every* subsequent user turn, creating a duplicate image placeholder block in the cache and biasing the model toward describing the image. Now it's only attached on the first turn after `/image`.
- **Cache divergence recovery.** For any residual tokenizer roundtrip gap (e.g. code blocks with mixed indent+newline tokens), trim the cache to the longest common prefix with the new prompt and re-prefill the divergent suffix — logs a `WARN` so the remaining roundtrip work remains visible.
- **Test.** ``tests/test_gemma4_audio_image_audio.cpp`` reproduces the bug end-to-end: feeds ``who_are_you.mp3`` (audio-only), then ``banner.jpg`` + ``"describe this"``, then ``2+2.mp3`` with the image still attached, and asserts turn 3's response differs from turn 2's.

## Test plan

- [x] ``tests/build/test_gemma4_audio_image_audio`` passes (turn 3 answers ``"2 plus 2 equals 4."``)
- [x] ``chat`` manually: primary audio → image+describe → audio sequence answers ``"4"`` 3/3 runs
- [x] ``chat`` manually: audio → image+describe → text follow-up still describes correctly
- [x] ``chat`` manually: audio-only multi-turn unchanged (`"Four."`)
- [x] ``chat`` manually: ``/clear`` between image and audio recovers cleanly
- [x] Stress: 4 turns alternating audio / image+text / audio / audio, no regressions
- [x] Code-example prompt (previously tripped a hard error) now recovers via partial cache trim + `WARN` log

Run locally:
```sh
CACTUS_TEST_GEMMA4_MODEL=/path/to/gemma-4-e2b-it \
CACTUS_TEST_REPO_ROOT=/path/to/repo \
./tests/build/test_gemma4_audio_image_audio
```

### DIFF
--- a/cactus/engine/engine.h
+++ b/cactus/engine/engine.h
@@ -768,6 +768,7 @@ public:
 
 
     void set_cache_window(size_t window_size, size_t sink_size = 4) { kv_cache_.set_window_size(window_size, sink_size); }
+    size_t get_cache_size() const { return kv_cache_.current_seq_len; }
 
     bool load_npu_prefill(const std::string& model_path);
     bool has_npu_prefill() const;
--- a/cactus/engine/engine_bpe.cpp
+++ b/cactus/engine/engine_bpe.cpp
@@ -156,6 +156,34 @@ bool BPETokenizer::load_vocabulary_mmap(const std::string& vocab_file, const std
             priority++;
         }
     }
+    
+    auto is_whitespace_only = [](const std::string& s) {
+        if (s.empty()) return false;
+        for (char c : s) {
+            if (c != ' ' && c != '\n' && c != '\t' && c != '\r') return false;
+        }
+        return true;
+    };
+    std::vector<std::string> ws_tokens;
+    for (const auto& tok : id_to_token_) {
+        if (tok.size() >= 2 && is_whitespace_only(tok)) ws_tokens.push_back(tok);
+    }
+    std::sort(ws_tokens.begin(), ws_tokens.end(),
+              [](const std::string& a, const std::string& b) { return a.size() < b.size(); });
+    for (const auto& tok : ws_tokens) {
+        for (size_t i = 1; i < tok.size(); i++) {
+            std::string first = tok.substr(0, i);
+            std::string second = tok.substr(i);
+            if (!token_to_id_.count(first) || !token_to_id_.count(second)) continue;
+            std::string key = first + "\x00" + second;
+            if (merge_map_.find(key) == merge_map_.end()) {
+                merge_rules_.emplace_back(first, second, tok, priority);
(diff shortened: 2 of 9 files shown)


kar-m:
lgtm @HenryNdubuaku

ncylich:
Thanks for the review — all three critiques were valid and I pushed fixes (force-pushed, still DCO-signed).

1. **`/image <same_path>` not reattaching.** Dropped the path-equality guard — `image_committed` is now unconditionally reset on any `/image` command, so retyping `/image banner.jpg` attaches again on the next turn.

2. **KV cache trim off-by-one.** You're right that `handle->processed_tokens` is usually one ahead of `kv_cache_.current_seq_len` (the last sampled token isn't forwarded yet), so `remove_token_range(common, processed_tokens.size() - common)` was hitting the `start+count > current_seq_len` early-exit in `engine_cache.cpp:479` and silently no-op'ing. Added a small public `Model::get_cache_size()` accessor and clamp the removal count to `kv_len - common`. Verified: on the code-block stress case the WARN now fires *and* the actual trim lands, so the model reaches the audio question correctly without relying on `decode_multimodal`'s secondary reset.

3. **`cactus_reset()` missing `user_audio_counts` clear.** Fixed in `cactus_init.cpp` — `cactus_reset()` now clears it alongside `processed_tokens` / `processed_images`.

Verified end-to-end with `tests/test_gemma4_audio_image_audio` plus chat manual runs (primary, `/clear`, `/image` retype, `reset`, code-example stress).

PR #607, opened 2026-04-23
Karen/opf

kar-m:


kar-m:
cactus download openai/privacy-filter --precision INT4
cactus build
CACTUS_OPF_MODEL=weights/privacy-filter tests/build/test_opf

PR #609, opened 2026-04-24
Sampler Bitmask v2

mhayes853:
Adds a bitmask parameter to sampling kernels/graph to enable structured generation.


ammesatyajit (approved):
reviewed on call with matthew, looks good

PR #624, opened 2026-05-04
Cross-framework INT8 op bench + hybrid INT8/FP16 decode kernel optimizations

ncylich:
## Summary

Four commits on top of \`v2\`:

1. **\`da99f6ce\` Add cross-framework int8 op benchmark suite** — five-graph deck (3 matmul shapes + prefill + hybrid decode) comparing cactus to ggml/llama.cpp, LiteRT, ExecuTorch, ONNX Runtime under matched precision and cold-cache pressure.

2. **\`851aa65f\` Trim non-essential comments from bench suite** — formatting only, 147 lines removed.

3. **\`727e7995\` Optimize hybrid INT8/FP16 decode attention kernel** — vdotq_s32 INT8 K dot path, deferred K scale (single fp32 mul/group via vmlaq_n_f32), 4-position K/V interleave, FP16 V block accumulator widened to FP32 every BLOCK_SIZE positions, light prefetch, stack scratch buffers. ~2× faster decode at every cache length, accuracy unchanged. Portable to any ARMv8.2-A+ \`+dotprod+fp16\` core.

4. **\`fe7b8b15\` Extend hybrid INT8/FP16 decode fast path to sliding-window attention** — the seq_len=1 fast path now handles windowed layers (Gemma-style 4:1 sliding/global pattern) by clamping kv iteration to \`[max(0, q_pos − window), q_pos+1)\`. Dispatcher gates the fast path on an unrolled cache (\`position_offset <= cache_len\`) so the streaming-LLM ring buffer's slot/abs divergence falls back safely to the general path. Bench gains \`--kv_heads\` / \`--head_dim\` flags for GQA/MQA configs; \`accuracy_test\` gains a rolled-cache regression sweep that locks in the invariant.

## Numbers

**Kernel-level (single-thread, default config q=kv=8, hd=128):**

| cache_len | v2 baseline | this branch | speedup |
|---|---|---|---|
| 128 | 31µs | 16µs | 1.94× |
| 256 | 52µs | 22µs | 2.36× |
| 512 | 102µs | 46µs | 2.22× |
| 1024 | 202µs | 90µs | 2.24× |
| 2048 | 400µs | 188µs | 2.13× |
| 4096 | 797µs | 416µs | 1.91× |

**End-to-end Gemma 4 E2B-it** (35 layers — 28 sliding window=512, 7 global; q=8, kv=1, hd=256), greedy decoding, max_tokens=512, 15 interleaved samples per kernel:

|  | this branch | v2 baseline |
|---|---|---|
| decode tok/s (mean) | 32.5 | 29.2 |
| decode tok/s (median) | 32.4 | 29.6 |
| stdev | 1.6 | 1.0 |

**Speedup: 1.10× (10%) [95% CI 7-13%], t = 6.2, p < 0.0001.**

## Accuracy

8 random seeds × 6 cache lengths × 3 regimes (no window / windowed unrolled / windowed rolled-cache):

| regime | max nrmse | pass |
|---|---|---|
| no window | 0.0043 | ✓ |
| sliding window=512, unrolled | 0.0043 | ✓ |
| sliding window=512, rolled (roll=128/512/2048) | 0.0041 | ✓ |

All under the 0.005 INT8 group=32 noise floor.

## Test plan

- [x] \`accuracy_test\` passes all three regimes locally
- [x] \`attn_bench --graphs attn_decode_cache --backends cactus\` matches the reported speedups across cache_len ∈ {128, 256, 512, 1024, 2048, 4096}
- [x] End-to-end Gemma 4 E2B-it with the engine's \`chat\` binary, greedy decode, 15 interleaved warm samples, p < 0.0001 vs v2 baseline
- [x] Output coherent (greedy generation produces non-degenerate text)
- [ ] Reviewer to spot-check the rolled-cache test reference vs the general-path masking semantics
- [ ] Reviewer to verify the dispatcher gate covers all paths (\`is_causal == false\`, prefill, GQA configurations)

### DIFF
--- a/cactus-kernels/src/attention.cpp
+++ b/cactus-kernels/src/attention.cpp
@@ -676,9 +676,9 @@ void cactus_attention_f16(
         });
 }
 
-// Decode-only specialization (seq_len=1, head_dim%32==0, quant_group_size=32, no
-// window, fully causal). Pre-quantizes Q to INT8 once per head, then uses
-// vdotq_s32 in K scoring — ~3× faster K dots than the FP16 FMA path.
+// Decode-only specialization. Pre-quantizes Q to INT8 once per head, then uses
+// vdotq_s32 in K scoring. Sliding-window only safe when the cache is unrolled
+// (slot index == absolute position); the dispatcher enforces this.
 static void cactus_attention_hybrid_int8_fp16_decode_dot(
     const __fp16* queries,
     const int8_t* keys_cached,
@@ -696,7 +696,8 @@ static void cactus_attention_hybrid_int8_fp16_decode_dot(
     size_t head_dim,
     float scale,
     size_t position_offset,
-    bool is_causal
+    bool is_causal,
+    size_t window_size
 ) {
     const size_t kv_seq_len = cache_len + new_len;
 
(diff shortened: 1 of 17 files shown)


ncylich:
Splitting into two MECE PRs: kernel-only and bench-only.

PR #660, opened 2026-05-21
Karen/v2 transcribe

kar-m:


kar-m:
added support for conversion of parakeet 1.1b tdt, previously it got hung up on nonstandard nemo packaging that differed from parakeet 0.6b; whisper models didnt work before because their encoder output wasnt statefully saved/passed forward, so the generation just died after the encoder ran successfully. Both fixed, tested parakeet 0.6b, 1.1b, whisper tiny, small, medium, large-v3-turbo;  RAM usage issues present but will be fixed by noah's PR

HenryNdubuaku:
@ParkiratS :)

PR #686, opened 2026-06-03
Long-context: rolling KeyDiff KV-cache compaction + chunked prefill

ncylich:
## Summary

Umbrella PR for the long-context work: bounded **rolling KV-cache compaction** that works across architectures, plus the chunked-prefill and RoPE-precision fixes it depends on. Rolling compaction is **default-on** for causal LMs — when the live KV cache reaches the trigger length it KeyDiff-compacts back to the target length (sink + recent + distinctive-middle) and renumbers, so the cache stays bounded through long decodes.

## What's included

**KeyDiff KV-cache compression** (`cactus-engine/src/kv_compress.{h,cpp}`)
- Query-agnostic per-head keep-set selection (sink + recent + distinctive-middle) for fp16 and int8 caches. Keys are un-RoPE'd before scoring (`s_i = -cos(k_i, mean(k))`) so the keep-set is position-frame-independent.
- Default-on rolling mode: trigger 4096 / target 2048, recent fraction 0.30, sink 4 — set via `kv_compress_trigger_len` / `kv_compress_target_len` (and `kv_compress_recent_frac` / `kv_compress_sink`), overridable at runtime with `CACTUS_KV_COMPRESS_AT` / `CACTUS_KV_COMPRESS_TO` (`AT=0` disables). Invalid rolling configs disable the flag instead of corrupting the cache.

**Sliding-window re-rope (pure engine, works on existing bundles)**
- At each compaction, global/full-attention layers compact + renumber to `0..B-1`; sliding-window layers rotate their recent keys in place by `-Δ` using the local RoPE theta. Since RoPE is relative, a single compressed position frame serves both — no graph/transpiler/manifest change. Enables Gemma-style sliding+global hybrids to compact correctly. KV-shared global source layers are re-roped with the global theta.

**Hybrid-cache safety (LFM2 etc.)**
- Compaction skips any cache state whose key node is not `KV_CACHE_STATE`, so conv/recurrent caches in hybrid models are left untouched instead of being misclassified as attention KV and corrupted.

**Special-token / thinking-token handling**
- Protected positions (special tokens) are always kept across compaction; the thinking-token cache strip is skipped on an already-compacted cache.

**On-demand KV cache + prefill handoff**
- The KV cache grows on demand and is moved across the prefill→decode handoff; caches (Qwen chunked-prefill, LFM2 decoder_step) are sized from the full context, with full-context cache transpilation support.

**Chunked prefill**
- Fix sliding-window KV-cache corruption from tail-chunk padding (`cactus-graph` window-size plumbing + the engine's chunked-prefill path).
- Default to the chunked-prefill component pipeline for text Qwen3 (~5x faster prefill, identical output); gated on the `Qwen3ForCausalLM` architecture so other families stay on the monolithic decoder.

**RoPE precision (long-context correctness)**
- Precompute cos/sin tables in the transpiler (`optimize_graph.precompute_rope_tables`) so positions past the fp16-angle range no longer garble; `max_position_embeddings` is carried in common graph meta to size the table.

## Testing
- `test_kv_compress.cpp`: keep-set selection, fp16/int8 compaction, int8 refactor bit-equivalence, sliding+global rolling over multiple cycles, local-theta load-bearing, no-op guards.
- `python/tests/test_optimize_gemma4_attention.py`, `test_cli_transpile_defaults.py`, `test_cli_run.py`, `test_bindings.py` for the transpiler/CLI wiring.
- `transpile/tools/benchmark_gemma4_context_scaling.py` for Gemma-4 context-scaling measurement.
- Live e2e: Qwen3 (all-global) and Gemma-4 (sliding+global) stay coherent across multiple in-decode compaction cycles; LFM2 hybrid coherent after the cache-type fix.


ncylich:
Rolled additional long-context work into this PR (3 commits on top of KeyDiff):
- **Precompute RoPE cos/sin tables** — fixes fp16 RoPE-angle corruption past ~6k context (Qwen3/Gemma now match the MLX reference through 16k). Decoder-gated; vision/audio encoder rope untouched.
- **Gemma chunked-prefill sliding-window fix** — tail-chunk padding was corrupting the sliding-window KV cache; deep-context retrieval 3/9 → 7/9 (≈ MLX).
- **Rolling KV compaction is now the default** for causal LLMs (4096→2048; `CACTUS_KV_COMPRESS_ROLL=<t>:<t>`/`off` overrides). One-shot mode removed. STT excluded structurally (separate transcribe API path); sliding-window models cleanly no-op for now (per-layer-RoPE-position support to follow so Gemma compacts its global layers).

PR #688, opened 2026-06-04
Long-context fixes: RoPE precision + Gemma chunked-prefill sliding-window cache

ncylich:
## Summary
Two independent long-context correctness fixes, each validated against the MLX reference (which stayed correct where the Cactus engine diverged).

### 1. RoPE precision (transpiler)
The transpiled graph computed the RoPE angle (`position * inv_freq`) in fp16, which cannot represent absolute positions > 2048 exactly. Past ~6k context the angle error reaches several radians and randomises `cos/sin`, corrupting generation. A new `optimize_graph` pass precomputes `cos/sin` offline in fp64, stores them as fp16 constant tables, and gathers them by position id — **no new kernels, no fp32 runtime path**. Gated to the text decoder (vision/audio encoder rope untouched); resolves each `cos/sin` node's own `inv_freq` so multi-theta models (Gemma sliding vs global) work.

**Before:** Qwen3/Gemma generation garbles past ~6k. **After:** matches MLX through 16k.

### 2. Gemma chunked-prefill sliding-window cache (engine)
`run_chunked_prefill` pads the final partial chunk to `chunk_size`; global layers trim the padding from the KV cache but sliding-window layers do not — the padding evicts real recent tokens and shifts the sliding window, corrupting deep-context attention. Fix: skip tail padding when the prefill graph has a sliding-window KV cache and process the tail token-by-token (already exact). Qwen (no sliding window) unaffected.

**Before:** Gemma 8k needle retrieval 3/9 (deep needles digit-garbled). **After:** 7/9 (≈ MLX's 8/9).

## Verification
- `test_precompute_rope_tables.py`: dual-theta fp64 table correctness, cat-duplication layout, position binding, Qwen regression, and vision/lm-encoder inertness (skips when transpiled bundles are absent).
- Coherence probe matches MLX `BLUE-42-FALCON` at 2k/8k/16k on Qwen3-1.7B and Gemma-4-e2b.
- Gemma NIAH 8k: depth-50 0/2 -> 2/2 exact; depths 10/50/90 = 7/9. Chunked-prefill speed retained (~200 prefill tps).

## Note
The RoPE table is sized to `max_cache_seq_len`. Absolute positions beyond it (rolling/streaming cache past the cache window) raise a fail-loud out-of-bounds error rather than silently corrupting — that regime produced garbage before this fix.

### DIFF
--- a/cactus-engine/src/model.cpp
+++ b/cactus-engine/src/model.cpp
@@ -1072,26 +1072,31 @@ Model::ChunkedPrefillResult Model::run_chunked_prefill(const std::vector<uint32_
     size_t effective_chunk = chunk_size > 0 ? std::min(chunk_size, component_tokens) : component_tokens;
     if (effective_chunk != component_tokens) effective_chunk = component_tokens;
     size_t whole_chunks_end = (tokens.size() / effective_chunk) * effective_chunk;
-    const bool has_recurrent_state = [&]() {
+    auto any_cache_node = [&](auto predicate) {
         if (!decoder_prefill_->graph) return false;
         for (const auto& state : decoder_prefill_->cache_states) {
             for (int node_id : {state.key_node_id, state.value_node_id}) {
                 if (node_id < 0) continue;
-                if (decoder_prefill_->graph->get_node_op_type(static_cast<size_t>(node_id))
-                    == OpType::RECURRENT_CACHE_STATE) {
-                    return true;
-                }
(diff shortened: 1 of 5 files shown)


ncylich:
Superseded by #686, which now carries these changes (RoPE precompute-table fix + Gemma chunked-prefill sliding-window fix) alongside the KeyDiff / rolling-KV-compaction work. Consolidating into #686 as the umbrella PR.

PR #694, opened 2026-06-07
Add Python inference benchmark CLI

RitwijParmar:
Adds a `cactus benchmark` command for repeatable local inference measurements from the Python CLI.

Why this felt useful: the Python API already returns TTFT, prefill throughput, decode throughput, RAM, and token counts, but there was not a small built-in way to run the same prompt profiles repeatedly and save artifacts for comparing runtime, quantization, or long-context changes.

What changed:
- adds built-in benchmark profiles for short chat, long prefill, and JSON-shaped output
- supports custom profile files so a PR or device run can publish exact prompt shapes
- writes per-run JSONL plus grouped summary JSON with mean, p50, p95, min, and max
- resets the KV cache between runs by default, with `--keep-cache` when cache reuse is intentional
- documents the workflow in the Python README

I used Codex to help implement and test this, and reviewed the code locally before opening the PR.

Tests:
- `/Users/ritwij/Documents/oss-work/cactus/.venv-codex/bin/python -m pytest python/tests/test_cli_benchmark.py python/tests/test_cli_run.py -q`
- `/Users/ritwij/Documents/oss-work/cactus/.venv-codex/bin/python -m pytest python/tests/test_bindings.py::TestCliParser python/tests/test_cli_benchmark.py -q`

Both passed locally before I removed the temporary test venv.

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -178,6 +178,32 @@ cactus benchmark \
   --markdown-report runs/regression_report.md
 ```
 
+Use a budget file when CI or device-lab runs need different tolerances per
+metric or profile. For example, a long-context sweep can use p95 decode
+throughput while RAM gets a tighter global budget:
+
+```json
+{
+  "threshold_pct": 5,
+  "metrics": {
+    "ram_usage_mb": {"threshold_pct": 2}
+  },
+  "profiles": {
+    "context_sweep_4096": {
+      "decode_tps": {"stat": "p95", "threshold_pct": 3},
+      "time_to_first_token_ms": {"threshold_pct": 8}
+    }
+  }
+}
+```
+
+```bash
+cactus benchmark \
+  --compare runs/main_summary.json runs/pr_summary.json \
+  --budget-json runs/mobile_budgets.json \
+  --fail-on-regression
+```
+
 ### Prefill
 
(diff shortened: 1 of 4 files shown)


RitwijParmar:
I pushed a follow-up that turns this into more of a regression harness rather than just a runner.

New pieces:
- synthetic context sweep profiles via `--sweep-token-counts`
- environment metadata in summaries so runs are easier to compare later
- Markdown reports for benchmark artifacts
- `--compare base.json candidate.json` with thresholded regression checks for TTFT, total latency, prefill TPS, and decode TPS
- compare mode avoids importing the native binding, so it can run anywhere the summary artifacts are available

I also added tests around the sweep generation, comparison logic, Markdown output, and parser paths.

RitwijParmar:
Rebased this on current main and resolved the CLI conflicts. The benchmark command now sits next to the newer run/list/test CLI changes instead of replacing them. I also checked the touched Python files with Python 3.11 compile. Local pytest is not available in this environment, so I could not run the focused pytest here.

RitwijParmar:
Closing this PR because it has been inactive for several months and is no longer merge-ready. Thanks for the consideration.

PR #697, opened 2026-06-09
SME2 acceleration for CQ4 matmul, prefill attention, and orthogonal embeddings

ncylich:
## Summary

Adds an ARM **SME2** backend for the hot CPU inference paths — quantized matmul (GEMV + GEMM), chunked-prefill attention, and orthogonal embedding dequantization — behind runtime feature dispatch with NEON fallback everywhere SME2 is absent or doesn't win.

**End-to-end on Gemma 4 E2B IT** (Apple M4 Pro, 1068-token prompt, 32 tokens, temperature 0, MTP off; 6-cycle alternating A/B vs the production NEON path in the same binary, cycle 1 discarded):

| metric | NEON (`CACTUS_QUANT_BACKEND=1`) | SME2 (auto) | delta |
|---|---|---|---|
| decode | 35.8 tok/s | 48.0 tok/s | **+34%** (worst cycle +28.6%) |
| TTFT | 3413 ms | 2249 ms | **−34%** |
| prefill | 319 tok/s | 477 tok/s | **+49%** |

### Quantized matmul (`cactus-kernels`)
- Two new SME2 translation units: `matmul_sme2.cpp` (streaming SMOPA leaves) and `matmul_sme2_gemv.cpp` (LUTI4/ZT0 in-engine 4-bit codebook expansion; pinned to `-O1` — Apple clang 17 mis-compiles vg1x4 ZA-array dots at `-O2/-O3` into a runtime SIGILL). Only these two files get `+sme2` flags (`armv9-a` deliberately avoided: it implies non-streaming SVE2, absent on Apple Silicon). Scalar stubs keep the build green on non-SME2 toolchains; base library flags unchanged.
- **Hybrid fused GEMV (M=1)**: one pool dispatch — phase A does parallel Hadamard + int8 activation quantization with a spin barrier, phase B dynamically steals 64-channel super-blocks: SME workers stream packed nibbles through the LUTI4 partials leaf (half the weight bytes, ZA never read back into the core), the remaining workers run the production NEON kernels on the native packed layouts. The main thread participates as SME worker 0 with a spin-join — the `wait_all` condition-variable sleep alone was a 20–30% tax on sub-100µs calls. Kernel-level vs production NEON at Gemma shapes: **down_proj 2.8×, o_proj 3.2×, gate_up 1.6×, q_proj 1.4×, lm_head 1.1×** (DRAM-bound). Auto-dispatch for `K*N ≥ 3M` elements.
- **Fused SME2 GEMM (M>1)**: (M-tile × super-block) work stealing over the packed weight cache, four ZA tiles in flight, raw int32 partials with NEON rescale — **2.4–3.6× vs NEON at every M** (K=N=1024).
- The orthogonal-rotation CQ4 lm_head maps onto the same hybrid through an exact virtual 128-wide-group view.
- The packed SME weight cache builds once per weight from row-major **or** production `INTERLEAVED_4ROW` layouts, byte-identical for both (tested).

### Prefill attention (`attention_hybrid.cpp`)
New SME2 path for the cached-int8 KV segment of chunked prefill (fully causally visible when `position_offset ≥ cache_len`): 16-query-row tiles, SMOPA QK with flat per-row/per-kv scales (group-scale ratios requantized into the int8 values so ZA accumulates the whole head_dim with one 4 KB readout per 64-kv block), vectorized masked softmax, then **u8×s8 USMOPA** AV with per-(row, v-group, block) P quantization (softmax probabilities are non-negative — unsigned doubles the resolution). The fp16 new-token segment continues through the existing flash-softmax loop seeded from the cached statistics; sliding-window + attention-sink masking matches the incumbent exactly. **4.6× (global) / 2.7× (sliding-window)** at Gemma shapes. `CACTUS_SME_ATTENTION=0` disables.

### Orthogonal embedding dequantization (both backends benefit)
`compute_embedding_node` was un-rotating each unique token row with a serial **scalar** K² matvec (~2.4M scalar FMAs per row at K=1536), repeated per per-layer embedding lookup per chunk — one of the largest prefill terms. The new batched path dedups indices and runs one parallel 8-wide NEON pass with identical fp32-accumulate math.

### Runtime / threading
- `cpu_has_sme2()` (Apple sysctl / Android HWCAP2) additionally requires streaming `svcntb() == 64`: all layouts are pinned to 512-bit SVL, so other SME2 implementations **fall back to NEON for correctness** until SVL-parametric variants exist.
- `CACTUS_QUANT_BACKEND=0/1/2` (auto / force NEON / force SME2) for in-binary A/B; `CACTUS_SME_GEMV_WORKERS`, `CACTUS_SME_PROF=1` knobs.
- Strided round-robin partitioning for coarse `parallel_for` items (the static splitter floors `work_per_thread` and dumps the remainder on the last worker — measured ~60% pool idle on attention tiles).
- `BufferDesc` move construction/assignment now transfers the SME cache.

### Docs
`cactus-kernels/docs/sme/` is the engineering notebook: status board, verified intrinsic/API notes, hardware findings (per-cluster in-order SME command queue, ZA readout costs, toolchain traps), dated experiment log, runnable probes (`working-examples/`), and research notes.

## Correctness / test plan

- **61 tests across 7 suites pass** (`cactus-kernels/test.sh`), including new:
  - forced-backend CQ1–4 GEMV/GEMM vs the fp32 oracle through the real dispatch, with M/N tile-tail cases;
  - production `INTERLEAVED_4ROW` fixtures built as the exact inverse of the shipped decoder; SME-cache layout invariance (row-major vs interleaved byte-identical);
  - orthogonal lm_head vs oracle; batched embedding rows vs the per-row reference on both layouts;
  - attention differential + double-precision oracle over 4 cases (global mid-prefill, rolled sliding window with sink exemption, zero-offset window, batch=2 with partial tiles/blocks). Worst-case attention relative error ~0.6–0.7% on adversarial uniform synthetic vs ~0.1% for the fp16 incumbent.
- End-to-end temperature-0 output verified coherent on both backends; SME-path engagement verified by symbol-level profiling.
- Benchmarks use alternating best-of runs (thermal drift exceeds kernel deltas).

## Notes for reviewers

- The attention path is the first SME path with real numeric divergence from NEON (int8 Q/K, u8 P quantization) — bounded and oracle-tested as above; all matmul SME paths produce integer partials identical to the NEON SDOT inner loop.
- Perf policies (worker splits, 3M-element dispatch threshold, queue model) are measured on M4 Pro; they are heuristics to re-validate on other SME2 silicon — the SVL gate keeps such devices correct (NEON) in the meantime.
- Known follow-ups documented in `docs/sme/debug-log.md`: eager-parallel SME cache build at model load (first-call lazy builds cost ~8.5 s on cold start, currently hidden by warmup), tail-sized prefill graph components, SME decode attention for long contexts.

### DIFF
--- a/cactus-graph/cactus_graph.h
+++ b/cactus-graph/cactus_graph.h
@@ -223,6 +223,21 @@ private:
     size_t round_up_size(size_t size) const;
 };
 
+// One-time SME2 packed-weight cache for a CQ buffer (see CactusQuantMatrix::expanded_sme).
+// Held by shared_ptr so buffer copies share it; built lazily at first matmul via
+// cactus_quant_build_sme_cache. ~K*N/2 bytes per weight matrix.
+struct CactusSmeCache {
+    std::vector<uint8_t> packed;
+    std::vector<float> norms;
+    // Orthogonal (lm_head) extras: K x K fp16 rotation transposed + per-row norms replicated
+    // across the K/128 virtual groups (see cactus_quant_orth_sme_gemv).
+    std::vector<__fp16> rot_t;
+    std::vector<__fp16> norms_rep;
+    uint32_t virt_ng = 0;
+    std::once_flag once;
+    bool ready = false;
+};
+
 struct BufferDesc {
     std::vector<size_t> shape;
     size_t total_size;
@@ -237,6 +252,7 @@ struct BufferDesc {
 
     void* activation_scales_data = nullptr;
     std::unique_ptr<char[]> owned_activation_scales;
+    mutable std::shared_ptr<CactusSmeCache> sme_cache;
     size_t num_rows_for_activation_scales = 0;
 
     BufferDesc();
@@ -265,7 +281,54 @@ struct BufferDesc {
     const __fp16* cq_rotation = nullptr;
     uint32_t cq_flags = 0;
 
+    // Lazily build the SME2 packed cache for this weight buffer (no-op on non-SME hardware or
+    // unsupported group sizes; orthogonal weights get a virtual-group view + transposed rotation
+    // for the SME lm_head path). Thread-safe; heavy work runs once.
+    void ensure_sme_cache() const {
+        if (sme_cache && sme_cache->ready) return;
+        if (!cpu_has_sme2()) return;
+        if (group_size <= 0 || (group_size % 32) != 0) return;
+        const uint32_t bits = PrecisionTraits::cq_bits(precision);
+        if (bits < 1 || bits > 4) return;
+        const bool orth = (cq_flags & CACTUS_QUANT_FLAG_ORTHOGONAL) != 0;
+        if (orth) {
+            // lm_head shape only: single group spanning K, fp16 rotation, row-major nibbles.
+            CactusQuantMatrix W = to_cq_matrix();
+            if (bits != 4 || !cq_rotation || W.num_groups != 1 || W.group_size != W.K ||
+                (W.K % 128) != 0 || (cq_flags & CACTUS_QUANT_FLAG_INTERLEAVED_4ROW)) return;
+        }
+        {
+            static std::mutex create_mu;
+            std::lock_guard<std::mutex> lk(create_mu);
+            if (!sme_cache) sme_cache = std::make_shared<CactusSmeCache>();
+        }
+        std::call_once(sme_cache->once, [&] {
+            CactusQuantMatrix W = to_cq_matrix();
+            if (orth) {
+                // Virtual 128-wide regrouping (exact: norms constant across subgroups).
+                const uint32_t vng = W.K / 128;
+                sme_cache->virt_ng = vng;
+                sme_cache->norms_rep.resize(static_cast<size_t>(W.N) * vng);
+                for (size_t n = 0; n < W.N; ++n)
+                    for (uint32_t g = 0; g < vng; ++g)
+                        sme_cache->norms_rep[n * vng + g] = cq_norms[n];
+                sme_cache->rot_t.resize(static_cast<size_t>(W.K) * W.K);
+                for (size_t k = 0; k < W.K; ++k)
+                    for (size_t i = 0; i < W.K; ++i)
+                        sme_cache->rot_t[i * W.K + k] = cq_rotation[k * W.K + i];
+                W.flags = 0; W.group_size = 128; W.num_groups = vng;
+                W.norms = sme_cache->norms_rep.data();
+            }
+            const size_t SB64 = (static_cast<size_t>(W.N) + 63) / 64;
+            sme_cache->packed.resize(SB64 * W.num_groups * static_cast<size_t>(W.group_size / 4) * 128);
+            sme_cache->norms.resize(SB64 * W.num_groups * 64);
+            cactus_quant_build_sme_cache(&W, sme_cache->packed.data(), sme_cache->norms.data());
+            sme_cache->ready = true;
+        });
+    }
+
     CactusQuantMatrix to_cq_matrix() const {
+        const bool sme_ready = sme_cache && sme_cache->ready;
         return CactusQuantMatrix{
             .bits = PrecisionTraits::cq_bits(precision),
             .K = static_cast<uint32_t>(shape.size() >= 2 ? shape[1] : shape[0]),
@@ -284,6 +347,8 @@ struct BufferDesc {
             .rotation = cq_rotation,
             .expanded = nullptr,
             .norm_f32 = nullptr,
+            .expanded_sme = sme_ready ? sme_cache->packed.data() : nullptr,
+            .norm_sme = sme_ready ? sme_cache->norms.data() : nullptr,
         };
     }
 
--- a/cactus-graph/cactus_graph.h
+++ b/cactus-graph/cactus_graph.h
@@ -280,13 +280,19 @@ struct BufferDesc {
     const uint32_t* cq_permutation = nullptr;
     const __fp16* cq_rotation = nullptr;
     uint32_t cq_flags = 0;
+    // True when external data points into a read-only file mapping (weights loaded via
+    // MappedFile) — required before releasing packed pages after the SME cache build.
+    bool mmap_backed = false;
+    // Stable per-weight-file identity (hash of resolved path; 0 = none). Graph components map
+    // the same weight file independently, so data pointers do NOT identify a weight.
+    uint64_t weight_key = 0;
 
     // Lazily build the SME2 packed cache for this weight buffer (no-op on non-SME hardware or
     // unsupported group sizes; orthogonal weights get a virtual-group view + transposed rotation
     // for the SME lm_head path). Thread-safe; heavy work runs once.
     void ensure_sme_cache() const {
         if (sme_cache && sme_cache->ready) return;
-        if (!cpu_has_sme2()) return;
+        if (!cactus_quant_sme_enabled()) return;   // also skips builds under forced-NEON A/B
         if (group_size <= 0 || (group_size % 32) != 0) return;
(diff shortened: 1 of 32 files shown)


ncylich:
Follow-up commit 9f30254: the SME orthogonal lm_head path now targets the production **INTERLEAVED_4ROW** format (row-major orthogonal bundles were a transpiler bug and are retired from the new code paths).

The virtual 128-wide group regrouping turns out to be byte-exact on the interleaved layout too (panel bytes are k-chunk-ordered, so the gs=128 re-view lands on contiguous 256-byte sub-panels) — no repacking; the existing interleaved preexpander and block kernels consume the virtual view natively. `ensure_sme_cache` now requires IL4ROW for orthogonal weights, the orth driver's NEON co-workers use the interleaved block processor, and the batched embedding dequant is interleaved-only (legacy row-major bundles fall back to the incumbent per-row path and should be recompiled).

Validation: 61/61 tests (orth test converted to an interleaved fixture against the fp32 oracle), SME engagement verified on a current production compile, temperature-0 output identical across backends. Isolated lm_head op: SME best 0.98 ms vs NEON interleaved best 1.20 ms (DRAM floor ~0.77 ms). Note for benchmark interpretation: the corrected format also speeds up the **NEON baseline** (~+14% decode — its old row-major orth lm_head cost ~4.8 ms/step vs 1.33 ms interleaved), so backend-vs-backend deltas measured on corrected bundles are smaller than the headline numbers above (which were measured against the buggy-format vintage), while absolute SME throughput is unchanged (~47.6 tok/s decode).

ncylich:
Follow-up commit 1c2077d: **single runtime weight format — no more 2× weight RAM on SME devices.**

Demand paging was keeping the packed `.weights` bytes resident (NEON co-workers and small-shape GEMVs streamed them every token) on top of the materialized SME cache. Now the SME cache is the one runtime format for both engines:

- New `cactus_quant_esme_gemv_blocks`: NEON SDOT directly over the cache layout (its expanded byte order *is* the classic `[4ch×4K]` SDOT panel). All tuned production-kernel techniques carried over; kernel-level parity-or-better vs the file kernels everywhere (pure-NEON small shape 1.29×, hybrid ratios unchanged).
- With a cache present, the fused driver is always the M=1 dispatch (SME share sized by shape, down to 0 = pure NEON over the cache); no file-layout matmul kernel runs in auto mode.
- Packed file pages are `madvise(MADV_DONTNEED)`'d after the one-time build (gated on a new `mmap_backed` flag; norms/codebook/rotation stay resident; forced-NEON A/B refaults cleanly). `CACTUS_QUANT_BACKEND=1` now skips cache builds entirely.
- Found and fixed a pre-existing 2× on top: every graph component maps the same weight file independently, so per-BufferDesc caches were built **once per component** (554 builds for 277 weights). A process-wide weak_ptr registry keyed by weight-file identity dedupes them.

**Measured (Gemma 4 E2B, auto backend): physical footprint at the same late-decode moment 4.9 GB → 3.7 GB (−1.2 GB); max RSS 5992 → 5018 MB.** Decode/TTFT at parity in alternating binary-vs-binary runs; 61/61 tests; temperature-0 output coherent on both backends. Known transient: the cache builder's int8 temp still spikes first-prefill peak (~400 MB on the lm_head) — streaming the build is a noted follow-up alongside eager build at load.

ncylich:
Follow-up commit 9e6ac5c: **re-measured everything against the same-format NEON baseline, and retuned the dispatch defaults accordingly.**

Three-way comparison (legacy file-layout NEON vs pure NEON over the cache vs SME hybrid; alternating, kernel-level + E2E on Gemma 4 E2B):

| config | prefill tok/s | decode tok/s | TTFT |
|---|---|---|---|
| legacy file-NEON (`CACTUS_QUANT_BACKEND=1`) | 368 | 42.0 | 2902 ms |
| pure NEON over the cache (no SME GEMV/attn) | 487 | 51.4 | 2192 ms |
| **shipped default** (NEON GEMVs + SME GEMM + SME attention) | ~510 | **~50–52** | **~2.0–2.2 s** |

Two findings worth highlighting:
1. The long-standing "NEON collapses on K-heavy GEMV" behavior was a **file-layout artifact**: o_proj jumps 57–67 → ~200–235 GF on the cache layout. With that fixed, the queue-limited SME GEMV leaf nets *negative* E2E under sustained decode (51.4 vs 44.8 tok/s) despite bursty per-shape kernel wins — so `CACTUS_SME_GEMV_WORKERS` now defaults to **0** (env/setter re-enables it for experiments and other SME2 silicon; forced-SME clamps to ≥1 so the leaf stays test-covered).
2. SME's durable wins are the **single cache format itself**, the **fused M>1 GEMM** (2.4–3.6×), and **prefill attention** (~100 ms TTFT in direct A/B) — all retained.

Net vs the legacy production baseline: **decode +19%, TTFT −27.5%, physical footprint −1.2 GB**, with 61/61 tests passing. The benchmark harness now reports all three columns with speedups quoted against the same-format NEON baseline.

ncylich:
Follow-up commit (thread budget): the low thread counts in the legacy kernels were **deliberate mobile/battery policy** — restored them in the fused drivers (GEMV `ceil(N/256)`, GEMM one thread per M-tile), with SME workers living *inside* the budget (replacing NEON workers, never adding threads; auto k=2 on ≥4-thread non-giant shapes, override via `CACTUS_SME_GEMV_WORKERS`).

Surprise result: **the budgeted configuration beats core saturation end-to-end** — decode 52.7 tok/s / TTFT ~1990 ms vs ~50–51 / 2100–2200 for the all-cores configs — and the intermittent run collapses seen under saturation disappear.

Final shipped numbers vs the legacy production baseline at the **same thread counts**: **decode 40.6 → 52.7 tok/s (+29.7%), TTFT 2875 → 1993 ms (−30.7%), prefill +44%, physical footprint −1.2 GB.** At matched budgets the SME GEMV leaf is kernel-positive on K-heavy shapes (+20–23%) but E2E-neutral on M4; the durable SME wins are the single cache format, the fused M>1 GEMM, and prefill attention. 61/61 tests.

ncylich:
Follow-up commit (power frontier): measured the decode **power/speed frontier** directly with `powermetrics` (decode-only windows, thermally-paired passes) over a thread-budget × SME-worker grid:

| config | decode | CPU power | energy/token |
|---|---|---|---|
| legacy budget, no SME | 48.7 tok/s | 21.0 W | 430 mJ |
| legacy budget, 1 SME worker | 50.0 tok/s | 18.3 W | 367 mJ |
| **half budget, 2 SME workers (shipped)** | 49.1 tok/s | **16.1 W** | **328 mJ** |

**SME workers strictly dominate zero-SME at every thread budget — faster and lower power.** The earlier "SME GEMV is E2E-neutral" reading came from measuring tok/s only; the leaf's value is on the power axis. Shipped defaults: half the legacy GEMV thread budget with a flat 2 SME workers (+0.8% speed, −24% energy/token vs the legacy budget without SME; TTFT unchanged; 61/61 tests). Both knobs (`CACTUS_GEMV_SB_PER_THREAD`, `CACTUS_SME_GEMV_WORKERS`) stay runtime-tunable for per-device retuning — the Galaxy-class SoC probed alongside (SVE2 + SME1, no SME2) runs the NEON-over-cache path and is unaffected.

ncylich:
Superseded by the two-PR split: #709 moves the packed-panel layout into the .cact weight files themselves (no runtime cache — eliminates the lazy build, the page-release machinery, and the 2× RAM transient by construction) with NEON kernels for all aarch64, and #710 stacks the SME2 streaming layer on top reading the mmap'd panels zero-copy. Together they land everything this branch contained, with the format now explicit per file (.cq4p.weights) and validated E2E on Gemma 4 E2B IT (cumulative vs legacy: prefill +35%, decode +21%, TTFT −26%, decode energy/token −18%).

PR #701, opened 2026-06-10
cactus-graph: op bounds-checks, IR-load validation & buffer-lifetime fixes

ncylich:
Reduction/slice axis-bounds (builder, ops_math), MoE buffer allocation top_k factor (ops_nn), ADD_CLIPPED FP32/broadcast correctness (ops_math), TopK/seq-len underflow guards (ops_sample), output-slice shape (ops_tensor), image shape/size checks (ops_image), alias-root buffer-lifetime fix (execute), FFI null-cache-pointer guard (graph_ffi), malformed-graph overflow/precision/CQ-layout validation (io).

### DIFF
--- a/cactus-graph/src/builder.cpp
+++ b/cactus-graph/src/builder.cpp
@@ -220,6 +220,15 @@ size_t CactusGraph::reduction_op(OpType op, size_t input, int axis) {
     if (axis == -1) {
         out = {1};
     } else {
+        if (buf.shape.empty()) {
+            throw std::runtime_error("Cannot reduce a scalar tensor along an axis");
+        }
+        if (axis < 0) {
+            axis += static_cast<int>(buf.shape.size());
+        }
+        if (axis < 0 || static_cast<size_t>(axis) >= buf.shape.size()) {
+            throw std::runtime_error("Invalid axis for reduction operation");
+        }
         out = buf.shape;
         out.erase(out.begin() + axis);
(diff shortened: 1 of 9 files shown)


kar-m:
@ncylich some fixes seem redundant
Cannot fire on any input:
- io.cpp align_offset overflow (offset > SIZE_MAX - padding). All three call sites pass an offset bounded by file_size_ (a real mmap size): line 1024 is a small header constant; line 1029's scales_end is bounded by the PR's own new scales_bytes_ <= file_size_ check executed two lines earlier. To reach the overflow you'd need a ~16-exabyte file. This one is pure dead defense-in-depth.

Reachable only via malformed/crafted input (never fire for well-formed models — "hypothetical corruption" guards):
- io.cpp prec_val > CQ4, scales extend beyond file, CQ mul_checked/add_checked overflow, footprint > scales_bytes_, and the underflow-safe data_offset_+byte_size_ rewrite — all reachable with a crafted/truncated weight file (group_size_, num_groups_, byte_size_ are raw header fields not otherwise bounded). Real hardening of the file parser.
- ops_sample seq_len==0 and k > feature_size; ops_nn topk-shape and num_experts <= total; ops_image positive-dims / ch <= 3 / total_size match; ops_math compute_reduce_node axis — all reachable via malformed IR (file-supplied params/shape through add_node), unreachable through the normal transpiler which emits consistent values.

PR #702, opened 2026-06-10
engine+graph: thread-safe last_error + JSON option-parse hardening

ncylich:
Make last_error_message thread_local (utils.h declaration, cactus_graph.h declaration, last_error.cpp definition must match for ODR) so concurrent FFI calls don't corrupt each other's error string. utils.h: add try_parse_json_uint to reject negative/overflowing size fields (top_k, max_tokens, tool_rag_top_k, cloud_timeout_ms) that std::stoul would wrap.

### DIFF
--- a/cactus-engine/src/utils.h
+++ b/cactus-engine/src/utils.h
@@ -86,7 +86,7 @@ struct CactusModelHandle {
     CactusModelHandle() : should_stop(false) {}
 };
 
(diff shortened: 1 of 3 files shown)


kar-m:
LGTM @jakmro @HenryNdubuaku 
Reproduced the bug on main and confirmed it doesnt happen anymore after the fix is introduced

PR #703, opened 2026-06-10
cactus-engine: runtime correctness & memory-safety

ncylich:
bpe.cpp: fix BPE merge-key delimiter (the "\x00" C-string was empty, collapsing keys). tokenizer.cpp: skip empty special tokens to avoid an infinite split loop; fall back to safe vision dims to avoid div-by-zero. constraints.cpp: treat '}' inside a tool-arg string as literal. gemma_tools.h: don't stall the number parser on a non-numeric char. engine.h+complete.cpp: add can_generate() and reject generate/prefill on decoder-less (embedding/encoder-only) models instead of segfaulting. engine_image.cpp: guard negative dims / signed-multiply overflow. index.cpp: handle MAP_FAILED state + query-dim checks. rag.cpp: bounded response writes + JSON-escape content/metadata. model_npu.cpp: correct FP32 element/byte accounting in the audio NPU copy.

### DIFF
--- a/cactus-engine/src/bpe.cpp
+++ b/cactus-engine/src/bpe.cpp
@@ -148,7 +148,7 @@ bool BPETokenizer::load_vocabulary_mmap(const std::string& vocab_file, const std
             std::string merged = first + second;
             merge_rules_.emplace_back(first, second, merged, priority);
 
-            std::string key = first + "\x00" + second;
+            std::string key = first + '\0' + second;
             auto it = merge_map_.find(key);
             if (it == merge_map_.end() || priority < it->second) {
                 merge_map_[key] = priority;
@@ -175,7 +175,7 @@ bool BPETokenizer::load_vocabulary_mmap(const std::string& vocab_file, const std
             std::string first = tok.substr(0, i);
             std::string second = tok.substr(i);
             if (!token_to_id_.count(first) || !token_to_id_.count(second)) continue;
-            std::string key = first + "\x00" + second;
+            std::string key = first + '\0' + second;
             if (merge_map_.find(key) == merge_map_.end()) {
                 merge_rules_.emplace_back(first, second, tok, priority);
                 merge_map_[key] = priority;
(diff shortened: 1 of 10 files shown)


kar-m:
@jakmro should be good to merge

ncylich:
@kar-m @jakmro I reproduced/verified each change on the real code before this lands — net: it's all reachable safety/correctness, nothing latent, and no regression. No code change from me; the diff is unchanged.

**Clear bugs (verified):**
- **bpe.cpp** — `first + "\x00" + second` uses the `const char*` `operator+`, which stops at the null terminator, so the separator was silently dropped. Demonstrated standalone: `"ab" + "\x00" + "c"` has length 3 (equals `"abc"`) and collides with `"a" + "\x00" + "bc"`; the `'\0'` char overload gives length 4 and distinct keys. Origin had merge-map key collisions → wrong BPE merges for any vocab where two different splits concatenate equally (ubiquitous in BPE).
- **model_npu.cpp** — the FP32 audio-encoder path resized the slot by `n * sizeof(__fp16)` but the conversion loop wrote `n` floats = `n * sizeof(float)` — a 2x heap overflow (e.g. grow 3000 bytes, write 6000). Fixed sizing keys off the actual precision.
- **index.cpp** — on remap failure the old code left `mapped_index_` pointing at an already-`munmap`-ed region (dangling use-after-unmap); now the new maps are committed only after both succeed, with a failed-state guard on subsequent ops, plus query-embedding dim validation. `test_index` passes 8/8 including the error-path/dim-mismatch cases.
- **rag.cpp** — unbounded `strcpy` into the response buffer (overflow on long results/errors) replaced with a bounds-checked write; doc metadata was interpolated raw into JSON (unescaped) and is now run through `escape_json_string`.
- **constraints.cpp** — a `}` token inside a tool-call string argument (e.g. `{"code":"f(){}"}`) was treated as a structural closing brace and truncated the call. Removed; braces are only counted outside strings now (verified the state machine — structural brace handling is intact).
- **complete.cpp + engine.h** — calling complete/prefill on an embedding/encoder-only model (no decoder) segfaulted through the C FFI; now returns a clean error (`can_generate()` = `decoder_ != nullptr`).

**Defensive guards (reachable via malformed/edge inputs):** empty special token → infinite `std::string::find` loop (tokenizer hang guard); non-positive image dimensions (engine_image); uninitialized Gemma4 vision params → degenerate patch calc (fallback defaults); zero-progress in number parsing on malformed tool JSON (gemma_tools).

Verified `test_index` builds and passes 8/8 (the [ERROR] log lines are the deliberately-triggered error-path assertions, including the new dim-mismatch guard).

PR #704, opened 2026-06-10
telemetry: JSON-escape records + bound the offline event cache

ncylich:
JSON-escape telemetry record fields and cap the offline event cache file so it can't grow unbounded. test_telemetry.cpp keeps the per-task record count under the cap so the no-loss assertion stays exact.

### DIFF
--- a/cactus-engine/src/telemetry_impl.cpp
+++ b/cactus-engine/src/telemetry_impl.cpp
@@ -1101,19 +1101,25 @@ static CloudSendResult send_batch_to_cloud(const std::vector<Event>& local, cons
 static const size_t kMaxCachedEventsPerFile = 1000;
 
 static void trim_cache_file(const std::string& file, size_t max_lines) {
-    std::vector<std::string> lines;
+    // Stream the file keeping only the last max_lines, so a pre-existing oversized cache cannot blow up memory.
+    std::deque<std::string> lines;
+    bool trimmed = false;
     {
         std::ifstream in(file);
         if (!in.is_open()) return;
         std::string line;
-        while (std::getline(in, line)) lines.push_back(line);
+        while (std::getline(in, line)) {
+            lines.push_back(std::move(line));
+            if (lines.size() > max_lines) {
+                lines.pop_front();
+                trimmed = true;
+            }
+        }
     }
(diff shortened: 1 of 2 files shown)


kar-m:
@jakmro this introduces a design choice to cap local cache to 1000 lines, which makes sense but is also something we maybe wanna talk about on the pod tmrw? 
Otherwise tests pass and lgtm

ncylich:
@kar-m @jakmro agreed it's worth a quick pod conversation — here's the precise behavior, rationale, and the knobs so the discussion has full context.

**What the cap does**
- Cache is one file per event type (`<type>.log`); each is capped at `kMaxCachedEventsPerFile = 1000` (single named constant).
- On overflow it keeps the **newest** 1000 and drops the oldest (FIFO eviction). Trim runs after each flush-to-cache, under `telemetry_mutex` — the same lock as cache load/clear — so the read/truncate/rewrite is race-free.
- Worst-case on-disk bound is about 1000 × (number of event types) — a handful of files, on the order of a few hundred KB.

**Why**
- Without a cap the offline cache grows unbounded whenever the device can't reach the cloud for a long stretch (airplane mode, no network, cloud down) — the cap makes worst-case disk usage deterministic.
- Drop-oldest retains the most recent telemetry (usually the most useful); the cost is that events older than the last ~1000-per-type are discarded if the device stays offline that long without a successful flush.

**Knobs (all cheap to change after the pod)**
- The number is one constant.
- The eviction policy (drop-oldest vs drop-newest, or a single global cap instead of per-file) is a one-function change in `trim_cache_file`.

**Verified locally**
- Escaping (the other half of the PR): the real `escape_json_string` turns `Pixel "6"\Pro` into `Pixel \"6\"\\Pro`, and escapes newline/tab/control bytes to their JSON escape sequences — preventing malformed payloads / JSON injection from device model/OS/app_id/project fields.
- Cap: 1500 cached events trim to exactly 1000, keeping the newest (event_500..event_1499); 1000 or fewer is left untouched.
- The race/no-loss test was lowered to 960 events (8×120) to stay under the cap so its exact no-loss assertion still holds.

No strong opinion on the exact number or policy — happy to change either to whatever the team lands on at the pod.

PR #705, opened 2026-06-10
transpiler: JAX attention causality + reduction precision, attention-output layout guard

ncylich:
Three transpiler correctness fixes, each reached and validated with a runnable repro on the real capture/transpile pipeline (this branch vs `origin/main`). Final state: **2 files, 4 lines.**

### Contributions (validated)

**1. JAX prefill `is_causal` gating** (`jax_semantic_rewrites.py`)
A maskless self-attention is treated as causal only when the component is a known causal-prefill (`decoder_prefill_chunk` / `decoder_media_step`). Previously *any* maskless self-attention was forced causal, so bidirectional/encoder self-attention captured via JAX was silently causal-masked.
*Validated:* a maskless JAX self-attention (no component set) transpiled to output matching the **causal** reference and differing from true full attention by ~0.27; with the gate it matches full attention (FP16 noise).

**2. JAX `min`/`max` kept in native FP32** (`jax_semantic_rewrites.py`)
`min`/`max` are no longer legalized through FP16 (cast→reduce→cast). FP16 saturates at ±65504, so attention-mask sentinels (`-1e4`, `finfo.min`) and large logits overflow and corrupt exact min/max selection. `sum`/`mean` still legalize to FP16 (averaging at the engine working precision). Also `.strip()` the component-meta string.
*Validated:* a `min`/`max` node on FP32 gets 0 precision-casts on this branch vs 2 (forced FP16) before; `sum` still gets the FP16 pair.

**3. Attention-output layout guard** (`optimize_graph.py`)
`normalize_attention_layouts` no longer folds an attention's output permute into the op when that attention output is itself a graph output — folding left `graph.outputs` referencing a value no longer produced by any node (a dangling output). Reachable for models that return a raw attention tensor (encoders / feature extractors).
*Validated:* with the guard the graph output is preserved; without it, the output value is folded away and dangling.

### Excluded after validation
The original `fuse_attention` mask-input preservation was **dropped** — proven unreachable and, where its path is reachable, cosmetic:
- `fuse_attention` runs post-canonicalization (op always `attention`) and `match_attention` sets q/k/v = `inputs[0:3]`, so a non-window masked node is always idempotent. Instrumenting the mask-preserving branch across the full test suite + repros gave **0 hits**.
- The only masked-node rewrite is **sliding-window**, where both origin and this branch correctly drop the mask input (the window is carried by `window_size`); the sole difference was a leftover `additive_mask=True` attr on origin. Kernel test (`window_size=2`, no mask, `additive_mask` True vs absent) → **identical output (0.0 diff)**, since with no mask tensor the flag has nothing to apply to.

A correct guard would be >3 lines for no observable gain, so it was removed rather than kept as dead code.

### Verification
`pytest python/tests` + `pytest python/cactus/convert/tests` → **264 passed, 1 skipped** (JAX capture tests included).

### DIFF
--- a/python/cactus/transpile/jax_semantic_rewrites.py
+++ b/python/cactus/transpile/jax_semantic_rewrites.py
@@ -685,7 +685,9 @@ def _rewrite_jax_prefill_attentions(graph: IRGraph) -> None:
         node = graph.nodes[node_id]
         if node.op not in {"permute", "view", "reshape"}:
             continue
-        is_decoder_step = str(graph.meta.get("component", "") or "").lower() == "decoder_step"
+        component = str(graph.meta.get("component", "") or "").strip().lower()
+        is_decoder_step = component == "decoder_step"
+        is_causal_prefill = component in {"decoder_prefill_chunk", "decoder_media_step"}
         if node.op == "permute" and tuple(int(dim) for dim in node.attrs.get("permutation", ())) != (0, 2, 1, 3):
             continue
         if node.op in {"view", "reshape"} and not is_decoder_step:
@@ -733,7 +735,7 @@ def _rewrite_jax_prefill_attentions(graph: IRGraph) -> None:
             additive_mask=additive_mask,
             scale=scale,
             output_layout="bthd",
-            is_causal=mask_id is None and is_self_attention,
+            is_causal=is_causal_prefill and mask_id is None and is_self_attention,
             rewritten_from="jax_prefill_attention",
         )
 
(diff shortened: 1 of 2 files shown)


kar-m:
@jakmro looks good i think

ncylich:
@kar-m @jakmro heads up — I force-pushed a tightened version of this PR after reproducing every change against `origin/main` on the real pipeline. Net: it's now 2 files / 4 lines, and I **dropped** one change as unreachable.

**Kept (each reproduced with numerics):**
- **`is_causal` gating** — a maskless self-attention with an unknown/non-prefill component was being forced causal. Repro: transpiled output matched the *causal* reference and differed from true full attention by ~0.27; with the gate it matches full attention.
- **`min`/`max` native FP32** — they were legalized through FP16; FP16 (±65504) overflows mask sentinels (`-1e4`, `finfo.min`) and large logits, corrupting exact selection. Now native FP32; `sum`/`mean` still FP16.
- **attention-output layout guard** — folding the output permute when the attention output is itself a graph output left `graph.outputs` dangling (value no longer produced). Now skipped.

**Dropped: the `fuse_attention` mask-input preservation.** I could not reach it: `fuse_attention` runs post-canonicalization so the node is always canonical `attention`, which makes it idempotent for any non-window masked node; the only masked rewrite is sliding-window (mask intentionally dropped, carried by `window_size`). Instrumenting that exact branch across the whole test suite + repros gave 0 hits. A correct defensive guard would be >3 lines, so I removed it rather than keep dead code.

`pytest python/tests` + `pytest python/cactus/convert/tests` → 264 passed, 1 skipped (JAX capture tests run).

PR #707, opened 2026-06-10
convert/CLI: conversion & tokenizer robustness

ncylich:
convert/cli.py: re-raise on duplicate checkpoint keys, handle Mapping state dicts, report GPTQ fallback, fix a replace() policy. cactus_adapters/tokenizer.py: handle sparse vocab IDs + Gemma name detection. handoff_probe.py: validate checkpoint tensor shapes and write actual dims instead of hardcoded ones. model_adapters/adapters.py: config-derived Gemma4 shared-KV cutoff + Parakeet conv layout. cli/compile.py: guard missing brew on macOS. cli/model.py: pass the HF token via env to the convert subcommand.

### DIFF
--- a/python/cactus/cli/compile.py
+++ b/python/cactus/cli/compile.py
@@ -46,6 +46,8 @@ def _detect_sdl2() -> tuple[list[str], list[str]]:
     is_darwin = platform.system() == "Darwin"
 
     if is_darwin:
+        if not check_command("brew"):
+            return [], []
         sdl2_check = subprocess.run(["brew", "list", "sdl2"], capture_output=True)
         if sdl2_check.returncode == 0:
             sdl2_prefix_result = subprocess.run(
--- a/python/cactus/cli/model.py
+++ b/python/cactus/cli/model.py
@@ -1,6 +1,7 @@
 """Model resolution, weight management, and bundle preparation."""
 from __future__ import annotations
 
+import os
 import shutil
 from dataclasses import dataclass
 from pathlib import Path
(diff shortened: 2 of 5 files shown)


python/cactus/convert/handoff_probe.py
```diff
@@ -109,11 +109,35 @@ def export_gemma4_handoff_probe(output_dir: str | Path, *, model_id: str | None
     if missing:
         raise RuntimeError(f"handoff probe checkpoint missing tensors: {', '.join(missing)}")
 
+    feat = state["norm.weight"].numel()
```
kar-m: this block seems redundant? Seems like you're grabbig shapes from state dict and then comparing to shapes inside state dict? Looking more into this

kar-m: this should be reverted i think

kar-m:
tested, lgtm once the block is reverted @jakmro 
EDIT: Noah pushed update reverting said block, now lgtm

python/cactus/convert/handoff_probe.py
ncylich: You're right — reverted. That block derived feat/th/h1/h2 from the state dict and then "validated" the other tensors against those same derived dims, so it was a circular self-consistency check: it can't catch a genuinely wrong-but-consistent probe, and the handoff probe is a fixed artifact with known dims anyway. Reverted the whole block back to the original hardcoded `struct.pack`/manifest — `handoff_probe.py` is now identical to main, and convert tests still pass (53/53). The rest of the PR (convert/cli, tokenizer, adapters, cli) is unchanged.

PR #708, opened 2026-06-10
bindings: JNI & React-Native FFI memory-safety

ncylich:
cactus_jni.cpp: clear pending Java exceptions after callback dispatch, reserve local-ref capacity for batched index_add, and clean up partially-acquired refs on null array elements. Cactus.swift: validate equal lengths and uniform embedding-row widths before calling native index_add to prevent out-of-bounds reads.

### DIFF
--- a/android/cactus_jni.cpp
+++ b/android/cactus_jni.cpp
@@ -29,6 +29,7 @@ static void token_callback_bridge(const char* token, uint32_t token_id, void* us
     if (!env) return;
     jstring jtoken = token ? env->NewStringUTF(token) : nullptr;
     env->CallVoidMethod(ctx->callback, ctx->method, jtoken, static_cast<jint>(token_id));
+    if (env->ExceptionCheck()) { env->ExceptionDescribe(); env->ExceptionClear(); }
     if (jtoken) env->DeleteLocalRef(jtoken);
     if (attached) ctx->jvm->DetachCurrentThread();
 }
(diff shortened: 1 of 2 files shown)


android/cactus_jni.cpp
```diff
@@ -374,30 +376,47 @@ Java_com_cactus_CactusJNI_nativeIndexAdd(JNIEnv* env, jobject, jlong handle,
     jstring* metaStrings = new jstring[count];
     jfloatArray* embArrays = new jfloatArray[count];
 
+    env->EnsureLocalCapacity(count * 3);
```
kar-m: seems redundant without return value checking, also OOM on 3 * count elements doesnt seem particularly likely

kar-m: Also doesn't hurt and i'm not sure if it hints to the environment to allocate that memory

kar-m:
even with the issue in comment i think this lgtm @jakmro 
I tested and the original bug reproduces and this doesnt create regressions as far as i could test. Don't have a android device to test directly though
EDIT: Noah pushed change fixing, now LGTM

android/cactus_jni.cpp
ncylich: Good catch on the unchecked return value. Rather than drop the call, I made it meaningful: `EnsureLocalCapacity` is now checked at all three batch sites (nativeIndexAdd, IndexAddBuffers, QueryResults). On failure it leaves a pending OutOfMemoryError, so we gate element acquisition on success — if it fails the loop is skipped and the existing `acquired < count` path clears the exception and returns -1, instead of making further JNI calls with a pending exception. Kept the call itself since (as you note) it's harmless and is a cheap headroom hint for large batches on stricter JNI impls where the local-ref table isn't auto-grown. Done in 8a0f33f4.

ncylich: Update: on reflection you're right that the reservation isn't really pulling its weight, so I removed it rather than adding a return-value check. This is Android-only JNI and ART grows the local-reference table dynamically, so the JNI 16-slot floor that `EnsureLocalCapacity` targets isn't a real limit on the actual platform — the call was effectively a no-op hint. Dropped all three `EnsureLocalCapacity` calls; the real fix here is the null-element partial-cleanup (via `acquired`) plus clearing pending exceptions after callback dispatch. (Rebased the branch to a single clean commit, so my earlier reply's SHA is superseded.)

PR #711, opened 2026-06-10
Karen/gemmma multiturn template

kar-m:


kar-m:
@jakmro here i fixed the chat template for lfm and qwen, fixed function call decoders for them, fixed multiturn gemma function call, and cleaned out some comments. Tested on gemma, lfm, qwen, needle for function calling

PR #710, opened 2026-06-10
SME2 streaming kernels over the packed-panel weight format

ncylich:
Second PR of the panel-format split (stacked on #709 — retarget to main once that merges). The SME2 layer consumes the file-borne panels zero-copy: the panel bytes ARE the streaming layout, so there is no cache builder, no madvise machinery, and no second resident weight copy. Non-SME2 silicon (and any SME2 implementation with SVL ≠ 512-bit) falls back to the #709 NEON panel path bit-for-bit.

## What's here

- **Streaming leaves** (`matmul_sme2_gemv.cpp`, `matmul_sme2.cpp`): LUTI4/ZT0 nibble-expanding GEMV + fused SMOPA GEMM reading mmap'd panels directly, and the prefill-attention QK/AV SMOPA passes. All leaves emit raw int32 partials; the fp32 rescale stays in NEON, so SME numerics are identical to the NEON path (locked by an agreement test).
- **Dispatch**: SME workers replace NEON workers *inside* the existing mobile thread budget (flat `k = min(2, nt−1)`, the measured power/speed frontier; `CACTUS_SME_GEMV_WORKERS` overrides). The fused SMOPA GEMM takes all M>1 panel matmuls. `CACTUS_QUANT_BACKEND`/`cactus_quant_set_backend`: 0 auto, 1 force NEON, 2 force the streaming leaves.
- **Gating**: `cpu_has_sme2()` requires streaming `svcntb() == 64` — every layout is pinned to 512-bit SVL; other SVLs fall back to NEON for correctness. The SME TUs build with `armv8.2-a+…+sme2` (armv9-a implies non-streaming SVE2 that Apple Silicon lacks) and the GEMV TU at `-O1` (clang mis-compiles the vg1x4 ZA dots at -O2+ → SIGILL).
- **Docs**: `cactus-kernels/docs/sme/` knowledge base (design rules, gotchas, verified working examples for every layout).

## Tests (TDD: suite written first, implemented to green)

`[sme2]` variants of every panel test through the real dispatch — CQ1–4 GEMV, GEMM incl. M/N tails, orthogonal lm_head, interleaved fixtures — plus a NEON↔SME agreement gate on identical fixtures and the SME prefill-attention differential. 35/35 matmul + 6/6 attention + full graph suite green.

[...]

PR #712, opened 2026-06-11
Fuse dispatch for the legacy interleaved CQ4 GEMV

ncylich:
Third mode of the panel-format campaign (stacked on #710, which stacks on #709): the panel GEMV driver's dispatch retrofitted onto the **unchanged legacy interleaved micro-kernel**, so legacy-format bundles get the modern scheduling with zero file-format change.

## What changes

`cactus_quant_4bit_gemv_interleaved` previously paid three passes per call: a pool-parallel Hadamard transform with a condition-variable join, a serial per-group int8 quantize on the main thread, then a second static-partition dispatch with another cv join. It now runs the panel driver's structure in one dispatch:

- phase A (per-group Hadamard + int8 quantize) stolen by workers behind a spin barrier
- phase B dynamically steals 16-block chunks (64 channels, same economics as panel super-blocks)
- main thread participates as worker 0 and spin-joins (no cv sleep)
- same thread budget, via a shared `cactus_quant_gemv_sb_per_thread()` (`CACTUS_GEMV_SB_PER_THREAD`)

The 7-op inner loop, activation quantization math, and numerics are unchanged. CQ4 `INTERLEAVED_4ROW` only (the production legacy format); 1–3-bit wrappers untouched.

## Tests

Existing IL oracle test (`matmul_cq4_il[file]`, N=192) only exercises the serial path; added `matmul_cq4_il_mt[file]` (N=4164 → 1041 blocks → 66 chunks incl. a 1-block tail) to engage the multi-thread fused worker, phase-A stealing, and 4-chunk grabs. 36/36 matmul tests pass.

## Results

M4 kernel level (Gemma shapes, alternating best-of-5):

| shape | IL old driver | IL fused driver | panel NEON |
|---|---|---|---|
| kv_proj 1×1536×512 | 44.7 GF | 158.2 GF | 125.3 GF |
| o_proj 1×2048×1536 | 59.2 | 188.4 | 170.7 |
| down 1×6144×1536 | 153.8 | 268.1 | 229.3 |
| lm_head 1×1536×262144 | 727.4 | 901.7 | 897.7 |

M4 E2E, legacy bundles (~1k-token prompt + 32 decode, alternating cycles, coherent output):

| model | prefill tok/s | decode tok/s |
|---|---|---|
| gemma-4-e2b-it | 264 → 302 (+14%) | 41.4 → 49.2 (+19%) |
| qwen3-1.7b | 279 → 279 | 36.8 → 52.3 (+42%) |
| lfm2-350m | 142 → 183 (+29%) | 137 → 173 (+26%) |

On-device, legacy bundles (#698 harness, 512+32 spec, 3 alternating rounds, devices powered, Thermal Status 0):

| device | model | prefill | decode |
|---|---|---|---|
| Samsung SM-S942U1 | gemma-4-e2b-it | 312 → 304 (par) | 24.7 → 26.2 (+6%) |
| | qwen3-1.7b | 98 → 99 (par) | 19.5 → 23.1 (+18%) |
| | lfm2-350m | 70 → 105 (+51%) | 66.4 → 101.7 (+53%) |
| Pixel 10a (Tensor G4) | gemma-4-e2b-it | 86 → 84 (par) | 10.5 → 10.5 (par) |
| | qwen3-1.7b | 34 → 34 (par) | 11.2 → 10.9 (par) |
| | lfm2-350m | 38 → 44 (+16%) | 36.3 → 40.8 (+12%) |

Dispatch overhead is a fixed per-call cost, so the win scales with call rate: largest for small models on fast cores, parity (no regression) where slow cores already sit at the DRAM roofline. Decode-side this mode matches or beats the panel path on every phone case measured (the IL inner loop is 7 ops/32 weights vs the panel's 9, which matters on Cortex-X4-class issue width); the panel format keeps its structural wins — M>1 GEMM without the per-call expansion tax, SME2 zero-copy, no runtime cache.

With this, the engine has three coherent modes: fused-IL NEON on legacy bundles (best phone decode), panel NEON (portable prefill/TTFT), panel + SME2 (M-class frontier).

### DIFF
--- a/cactus-kernels/src/matmul.cpp
+++ b/cactus-kernels/src/matmul.cpp
@@ -1694,6 +1694,15 @@ static void cactus_quant_panel_rescale_sb(const int32_t* psb, size_t sb, const f
     for (uint32_t c = 0; c < valid; ++c) C[n_start + c] = static_cast<__fp16>(tmp[c]);
 }
 
+static size_t cactus_quant_gemv_sb_per_thread() {
+    static const size_t v = [] {
+        const char* e = getenv("CACTUS_GEMV_SB_PER_THREAD");
+        const int i = e ? atoi(e) : 8;
+        return static_cast<size_t>(i > 0 ? i : 8);
+    }();
+    return v;
+}
+
 // SME workers live INSIDE the thread budget, replacing NEON workers (measured frontier: flat k=2
 // dominates k=0 on speed AND power). Env/setter overrides; backend 2 clamps >= 1 for leaf coverage.
 static inline size_t cactus_quant_panel_k_sme(size_t nt, uint32_t gs) {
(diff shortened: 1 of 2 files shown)


ncylich:
Consolidating the stack into a single PR per review preference — this work now lives in #709.

PR #713, opened 2026-06-11
fixed needle rope bug, moved repeated rope table instantiations from …

kar-m:
…components into weights


kar-m:
@jakmro if you look at current model bundles for qwen, lfm, and gemma, they are pretty big outside of the weights because rope tables were being stored in the component folders. Some of these tables were being stored in several copies (gemma being the worst offender, storing 1.2gb worth of rope tables). Changed it so unique rope tables are stored in weights and just reused between components

PR #714, opened 2026-06-11
added language model embed component

kar-m:


kar-m:
@jakmro added embedding capability for langauge models, tested on lfm, qwen, gemma4. Haven't tested gemma3
validated against hf 
Essentially jsut added an extra component that runs like chunked prefill but doesnt include lmhead (some extra transpiler stuff for gemma)

PR #715, opened 2026-06-11
Simplify CLI: one bundle-resolution ladder + CQ level in list

rshemet:
## Summary

Two independent CLI cleanups (one commit each).

### 1. Unify model resolution behind one fallback ladder
`run`, `serve`, `transcribe` and `download` previously resolved bundles three different ways — `run` had the full ladder, `transcribe` skipped the cache/prebuilt rungs, and `serve` just errored telling you to run `convert` then `transpile`. They now share a single `ensure_runnable_bundle` resolver:

```
local bundle path -> cached build -> prebuilt HF download -> local convert+transpile
```

So every entry point builds a model on demand. `convert`/`transpile` are unchanged but demoted to an "Advanced / build pipeline" section in `--help`, since they now run automatically. Net diff is ~neutral — the shared resolver replaces three bespoke ones.

### 2. Show CQ quantization level in `cactus list`
Bit-width isn't stored in any single config field — it lives per-tensor in the `CACT` header. `list` now reads the precision code from `.weights` headers and reports the dominant CQ level as a column:

```
gemma-4   gemma4   CQ4   3.5 GB   .../weights
```

Early-stops after a small sample, so it stays fast (~0.2s).

## Test plan
- `cactus --help` renders; advanced section present
- `cactus list` shows CQ column
- all CLI modules import; parser still wires `convert`/`transpile`


kar-m:
tested, LGTM @jakmro

kar-m (dismissed):
LGTM (needed to write this in a review as well)

kar-m (approved):
the PR removed a function that was used in a test, after this was fixed the rest of the logic is good

PR #724, opened 2026-06-15
 Removed unnecessary null pointer checks

### LINKED ISSUE #667: Remove unnecessary null pointer checks
:eyes: [An extra null pointer check is not needed in functions](https://isocpp.org/wiki/faq/freestore-mgmt#delete-handles-null "Do I need to check for null before delete p?") like the following.
* https://github.com/cactus-compute/cactus/blob/3227a7f93083c9050b80c0d2b454e3302118c170/cactus/graph/graph_core.cpp#L71-L76
* https://github.com/cactus-compute/cactus/blob/3227a7f93083c9050b80c0d2b454e3302118c170/cactus/graph/graph_core.cpp#L108-L112

wasup-yash:
Fixes #667 

summary : 
It removes unnecessary null pointer checks, which were if conditions.

purpose of raising a PR:
The issue explains the reasoning behind it.

### DIFF
--- a/cactus-graph/src/core.cpp
+++ b/cactus-graph/src/core.cpp
@@ -69,10 +69,9 @@ BufferDesc::BufferDesc(const std::vector<size_t>& s, Precision prec)
 }
 
 BufferDesc::~BufferDesc() {
-    if (pooled_data) {
         delete[] pooled_data;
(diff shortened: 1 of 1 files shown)


elfring:
Thanks for your small source code improvement.

PR #736, opened 2026-06-17
fix(graph): apply SIGMOID/TANH activations in MoE layer

yus100:
The MoE layer ignored the `SIGMOID` and `TANH` activations. Both are valid options on the `moe_layer` builder, but the activation switch had no cases for them, so they silently fell back to SILU and produced wrong results.

This adds the two missing cases (the kernels `cactus_sigmoid_f16` and `cactus_tanh_f16` already existed). I also added a test for the MoE layer, which previously had none — it runs each activation through a tiny identity-weight MoE and checks the output matches the standalone activation kernel.

Confirmed the test fails before the fix and passes after. Run it with `cactus test graph --suite nn`.


HenryNdubuaku:
Thanks for this, i have assigned a reviewer.

cattermelon1234:
nice work stanley

yus100:
@kar-m

Issue #216, opened 2025-11-20
Any plans for supporting diffusion models?


adi776borate:
Seems like image generation models are not widely supported yet. Any plans for doing so? 
How feasible it will be?

HenryNdubuaku:
@adi776borate this is now in progress

adi776borate:
Hi! Could you please share the related docs? I tried searching through changelogs and commits but didn't find any references.

PR #779, opened 2026-08-08
Stable Diffusion 1.5

nshejwalkar:
This PR adds Stable Diffusion 1.5 (LCM) end to end. This is the first feature with text->image functionality in Cactus, as requested in [issue 216](https://github.com/cactus-compute/cactus/issues/216). It implements a pipeline of three models: a [CLIP](https://huggingface.co/openai/clip-vit-large-patch14) encoder, [U-net](https://huggingface.co/SimianLuo/LCM_Dreamshaper_v7/tree/main/unet), and [Tiny AutoEncoder for Stable Diffusion](https://github.com/madebyollin/taesd) as the decoder.

### cactus-kernels + cactus-graph
- Added upsample_nearest2d: a new graph op with NEON and Metal kernels.
- Added conv2d_k3s1p1: Metal kernel, 6.3× over the generic conv on TAESD decode.
- Exported both through the graph FFI so the transpiler can emit them.

### cactus-engine
- Added `cactus_generate_image` to the public C API, and a new translation unit that drives the pipeline: encode the prompt, run the denoising loop, decode the final latents, and pack the result as an RGB image.
- Added an `iterative_denoise` execution route.
- Implemented the LCM sampler on the host — timestep schedule, alpha table, and guidance embedding. The three graphs are pure tensor functions; all of the loop control lives here, because the original pipeline has Python between each stage.
- Scheduler settings (beta schedule, training timesteps, timestep scaling) are read from the bundle manifest, carried through from the model's own `scheduler_config.json` rather than hardcoded.
- Added CLIP's tokenization to the BPE tokenizer, which needs its own word splitting, normalizer, and decoder.

### python/convert
- Added support for diffusers checkpoints. These differ from transformers repos in three ways that all had to be handled: components live in separate subfolders, the weights file has a different name, and the model classes can't be constructed by `AutoModel` at all.
- Added family adapters for TAESD, the SD U-Net, and the CLIP text encoder. They keep the full checkpoint tensor name, so parallel stacks like `encoder.*` and `decoder.*` can't collapse onto the same output file.
- Conversion now fails loudly when two tensors would be written to the same filename. Previously the second one silently overwrote the first for TAESD.
- Added a CLIP tokenizer converter that emits the engine's sidecar files from the model's own tokenizer files.

### python/transpiler
- Added an SD1.5 profile declaring the three components, how they're wired, and that the runtime should drive them as a denoising loop.
- This is the first Cactus model that is a pipeline of independent models rather than a single checkpoint. Existing multi-component models like LFM2-VL are one checkpoint exported once and then sliced into components afterwards; here, each component is its own checkpoint with its own inputs, and the decoder comes from a different repository altogether. Added `ComponentSource` so a profile can say that: each part is loaded and exported on its own, reusing the existing loading and sample-input machinery rather than introducing a parallel path for diffusion.
- Mapped `upsample_nearest2d` through the IR and lowering, and taught the convolution fusions to tell 1D from 2D by kernel rank.
- Two export fixes: CLIP registers position ids as a buffer no checkpoint can supply, so it's recomputed in the graph; and attention is kept whole through decomposition so the existing attention fusion still matches (improved decode from 33s -> 8.2s).

### python/cli
- Added `cactus generate-image`, which prepares the bundle, generates, and writes a PNG.
- `cactus convert` now handles pipeline models, converting each component into a single shared weights directory. One component (the TAESD decoder) comes from a different repository entirely.
- Added the Python binding for the new image generation API.

### Tests (correctness and performance)
- kernels — upsample against a hand-computed reference, covering all three code paths in the kernel, plus benchmarks for the new convolution and upsample.
- graph — save/load round trip proving the scale factor survives serialization, and Metal/CPU parity.
- engine — unit tests for the sampler math (timestep schedule, alpha table, guidance embedding), including empty and edge-case inputs.
- convert — adapter selection and tensor naming, including that encoder and decoder names stay distinct.
- transpiler — each component transpiled and compared against PyTorch (CLIP 0.30%, U-Net 0.02%, decoder 0.17% relative error), plus the full text→image pipeline end to end.

### Example
```
cactus convert SimianLuo/LCM_Dreamshaper_v7
cactus generate-image --prompt "a red fox in snow" --out fox.png
```
fox

512×512 on M1 Pro w/ Metal: 8.2 s at 4 steps, 4.0 s at 2 steps.

Passes all correctness tests.


nshejwalkar:
The failed ci python test is because of e9773b1, unrelated to this PR

PR #784, opened 2026-08-12
Fix per-matrix CQ4 reciprocals in Metal batch transform

### LINKED ISSUE #773: gemma-4-E2B-it: current default cq4.zip (2026-07-09 "calibrated") collapses on multi-step reasoning — GSM8K 3/100 vs 87/100 for the previous build, at identical speed
### Summary

The current default download for `google/gemma-4-E2B-it` — `Cactus-Compute/gemma-4-E2B-it` `gemma-4-e2b-it-cq4.zip`, the file `cactus run`/`cactus download` resolve at `--bits 4` — appears to have lost multi-step reasoning in the 2026-07-09 artifact swap. The build it replaced (now `gemma-4-e2b-it-cq4-uncalibrated.zip` in the same repo) does not have the problem, and the two are speed-identical, so this looks like an unintended artifact regression rather than a speed/quality trade.

### Measurements

GSM8K, first 100 test questions, 0-shot CoT, greedy (`temperature 0`, `top_k 1`), `max_tokens 1024`, driven through `cactus_complete` (Metal backend, `auto_handoff: false`), same prompt + answer extraction for every build:

| bundle (sha256 prefix) | GSM8K |
|---|---:|
| `cq4.zip` (current default, `7a2579ad1089f626…`) | **3/100** |
| `cq4-uncalibrated.zip` (`7c71f075554bd27b…`) | **87/100** |
| `cq3.26.zip` (n=25 probe) | 1/25 |
| `cq2.54.zip` (n=25 probe) | 4/25 |

Supporting observations:

- The two cq4 builds decode at the same speed on an iPhone 17 Pro (50.2 vs 50.6 tok/s, median of 3 cold runs each) — the difference is purely output quality.
- The default build's failures are visible even on single-step questions: in a 9-item sanity prompt it answers 7×8 with "119" and "thank you in French" with "Bonjour"; the uncalibrated build gets 8/8 of its emitted items right.
- On multi-step problems the default build hedges without committing to a reading until the token budget runs out (~4.4k chars mean), the same signature across all 100 questions.
- Your docs (`docs/cactus_quants.md`) report GSM8K 71.20 for CQ4 and 74.00 for the production bundle — consistent with what we measure on the **uncalibrated** lineage, not with the current default.
- Timeline from the HF repo history: 2026-07-09 "Add locally-built calibrated CQ4 (GPTQ language, FP16 towers)" + renames made it the default; the 07-16/17 probe additions kept those graphs ("Restore original cq4 graphs; keep cloud-handoff probe").

### Repro

Everything is public in https://github.com/john-rocky/apple-silicon-llm-bench:

- harness: `scripts/parity_gsm8k.py` (`--which cactus`, drives your Python FFI; pinned question set in `evaldata/gsm8k_test.jsonl`)
- per-question reports: `results/quality/gsm8k_cactus-*.json`
- session audit trail incl. device runs: `results/raw/2026-07-20-cactus-arm/SUMMARY.txt`

```bash
python3 scripts/parity_gsm8k.py --which cactus \
  --bundle <unzipped gemma-4-e2b-it-cq4> --cactus-repo <cactus clone> \
  --n 100 --max-tokens 1024
```

Engine commit `1ace6d78`; results are deterministic under greedy decoding.

If the calibrated swap is intentional, a docs note would help users pick; if not, re-pointing the default at the uncalibrated graphs looks like the smallest fix. Happy to re-run the harness against a fixed artifact.

clumsylad21:
Fixes #773

## Summary

Fix the Metal CQ4 batched transform to use each matrix's own
`input_scale_recip` instead of reusing the first matrix's reciprocal
buffer across the batch.

Also adds a regression test covering batched CQ4 projections with
distinct reciprocal scales.

## Root cause

`cactus_metal_encode_transform_batch()` bound only `Ws[0]`'s
`input_scale_recip`, while `left_signs`, `right_signs`, and
`permutation` were already bound separately for each matrix.

As a result, `cq4_transform_batch` applied the first matrix's
reciprocal scaling to every projection in the batch.

This was effectively masked for the uncalibrated CQ4 weights, whose
input scales are unit-valued, but breaks calibrated weights where the
per-matrix scales differ.

## Validation

The targeted Metal regression test now passes:

    ✓ PASS │ matmul_cq4_metal_distinct_recip

The test uses three CQ4 matrices with deliberately distinct reciprocal
scales and checks the batched Metal path against independent Metal
execution.

I also reproduced the original model-level failure on the current
Cactus main baseline (`c90a1af`) using the calibrated Gemma CQ4 bundle.

Before the fix:

    pred=16, gold=18, chars=4350

With only the per-matrix reciprocal fix:

    pred=18, gold=18, chars=1018

The uncalibrated/calibrated CPU runs did not show this backend-specific
regression.

### DIFF
--- a/cactus-kernels/src/cactus_kernels.metal
+++ b/cactus-kernels/src/cactus_kernels.metal
@@ -138,16 +138,19 @@ kernel void cq4_transform_simd(
 
 kernel void cq4_transform_batch(
     device const half*  x       [[buffer(0)]],
-    device const half*  recip   [[buffer(1)]],
-    device const char*  lsign0  [[buffer(2)]], device const char* lsign1 [[buffer(3)]], device const char* lsign2 [[buffer(4)]],
-    device const char*  rsign0  [[buffer(5)]], device const char* rsign1 [[buffer(6)]], device const char* rsign2 [[buffer(7)]],
-    device const uint*  perm0   [[buffer(8)]], device const uint* perm1  [[buffer(9)]], device const uint* perm2  [[buffer(10)]],
-    device       half*  code0   [[buffer(11)]], device half* code1 [[buffer(12)]], device half* code2 [[buffer(13)]],
-    constant uint& ng           [[buffer(14)]],
+    device const half*  recip0  [[buffer(1)]],
+    device const half*  recip1  [[buffer(2)]],
+    device const half*  recip2  [[buffer(3)]],
+    device const char*  lsign0  [[buffer(4)]], device const char* lsign1 [[buffer(5)]], device const char* lsign2 [[buffer(6)]],
+    device const char*  rsign0  [[buffer(7)]], device const char* rsign1 [[buffer(8)]], device const char* rsign2 [[buffer(9)]],
+    device const uint*  perm0   [[buffer(10)]], device const uint* perm1 [[buffer(11)]], device const uint* perm2 [[buffer(12)]],
+    device       half*  code0   [[buffer(13)]], device half* code1 [[buffer(14)]], device half* code2 [[buffer(15)]],
+    constant uint& ng           [[buffer(16)]],
     uint tgpos [[threadgroup_position_in_grid]], uint lane [[thread_index_in_simdgroup]],
     threadgroup float* zmem     [[threadgroup(0)]])
 {
     uint g = tgpos % ng, b = tgpos / ng;
+    device const half* recip = (b==0u)?recip0:((b==1u)?recip1:recip2);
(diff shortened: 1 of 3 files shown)


jakmro:
Hey @clumsylad21, thanks for this PR, it just needs a [DCO](https://github.com/cactus-compute/cactus/pull/784/checks?check_run_id=94147540839) sign off

clumsylad21:
Thanks @jakmro ! Added the DCO sign-off.

PR #800, opened 2026-08-29
fix(kernels): route the remaining dot-product call sites through the non-DOTPROD fallback

avionicharshit-byte:
towards #561 , this does not close it , see the bottom

`matmul.cpp` already had a non-DOTPROD fallback , `cactus_dotq_lane0..3` behind `CACTUS_DOTQ_LANE` , used in 52 places. but 28 other call sites in the same file still call `vdotq_laneq_s32` directly , and `attention_hybrid.cpp` calls `vdotq_s32` in 10 places with no guard at all. so on a target without dotprod the code does not fall back , it fails to compile.

**what this does**

moves the guard block into `cactus-kernels/src/neon_dotprod.h` so both files share one switch , adds `CACTUS_DOTQ` for the non lane form which had no fallback at all , and routes the 38 raw sites through the macros. nothing changes on a dotprod target , the macros expand to the same intrinsics.

**testing , on an m-series mac**

- `-march=armv8.2-a+fp16` , no dotprod. before: 32 errors in `matmul.cpp` , 10 in `attention_hybrid.cpp`. after: 0 and 0
- bit exact , 200k random vectors through both paths across all 5 call forms , 0 mismatches. its integer maths so exact , not approximate
- `./test.sh` passes 50/50 native , and 50/50 again with the `#undef __ARM_FEATURE_DOTPROD` line uncommented so every site is forced down the fallback

**the loss**

the fallback is slower , measured with that switch on:

```
matmul_cq1 1024^3         2.818ms -> 17.781ms
matmul_cq4 1024^3         3.182ms -> 18.179ms
matmul_cq1 1x1024x1024    0.028ms -> 0.047ms
```

so about 6x on large gemm and 1.7x on the gemv shape decode uses. its a correctness path , not a speed path , and it only runs where the fast one would have been an illegal instruction.

**what this does not fix**

pi 4 still will not build. the `__fp16` arithmetic is a separate armv8.2 feature and has no fallback anywhere , with this branch applied its still 108 errors at `-march=armv8-a` across 11 files , all `fullfp16`. happy to do that next , either folded into this pr or as a second one.

one thing , i put the header at `cactus-kernels/src/neon_dotprod.h` , contributing says one header per folder so say the word and ill fold it into `cactus_kernels.h` instead.

### DIFF
--- a/cactus-kernels/src/attention_hybrid.cpp
+++ b/cactus-kernels/src/attention_hybrid.cpp
@@ -1,5 +1,6 @@
 #include "../cactus_kernels.h"
 #include "threading.h"
+#include "neon_dotprod.h"
 #include <arm_neon.h>
 #include <cmath>
 #include <algorithm>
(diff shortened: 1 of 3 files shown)


avionicharshit-byte:
@HenryNdubuaku quick question before i put more work into this

this pr is only the dotprod half , so on its own a pi 4 still wont build. the other half is the `__fp16` arithmetic , thats 108 errors across 11 files at `-march=armv8-a` with this branch applied , guarded by `__ARM_FEATURE_FP16_VECTOR_ARITHMETIC` and `__ARM_FEATURE_FP16_SCALAR_ARITHMETIC`

i can do that fallback the same shape as this one , convert to f32 and back only when those macros are absent , and again without touching any build flags , so nothing changes for chips that already have fp16

do you want it folded into this pr , or a second one stacked on top , or would you rather not carry a slow path at all

Udit8348:
For now, do not prioritize any work to bring up targets below ARM v8.2+
These processors are not targeted for "neural" workloads, in a way that cactus hw optimizations can take advantage of. (ie: fp16 support) There are obviously methods to emulate and add  legacy processor support, similar to how PyTorch does. On the other hand, this is not priority in my opinion and just clouds the src size, testing support, etc.

For now, it seems like build is configured to fail on unsupported architectures (https://github.com/cactus-compute/cactus/pull/768) This is sufficient observability for future debugging. IMO its better to first gauge audience interest with some kind of questionnaire in the readme.

avionicharshit-byte:
tested on an actual pi 4 , not cross compiled

pi 4 model b rev 1.4 , cortex-a72 , aarch64 , debian 13 , gcc 14.2. `/proc/cpuinfo` features are `fp asimd evtstrm crc32 cpuid` , so no `asimddp` and no `asimdhp` , a true armv8.0 target and the same board as #561

**upstream main on this pi** compiles fine with the current `-march=armv8.2-a+fp16+simd+dotprod+i8mm` and then dies at runtime. `cactus-kernels/test.sh` gives `Illegal instruction` on 6 of the 7 test binaries , 1 suite finishes , 13 of the 50 tests get to run at all

**at an honest armv8.0 baseline** , `-march=armv8-a+simd` , upstream does not compile , 118 errors across 11 files

**with this branch** , 88 errors , same 11 files , and every remaining one is `__fp16`. isolating the dotprod half with `-march=armv8.2-a+fp16` , fp16 on and dotprod off:

```
                       upstream   this branch
matmul.cpp                   16             0
attention_hybrid.cpp         30             0
```

what i cannot show yet is the suite passing on the pi , it still wont build there until `__fp16` arithmetic has a fallback too. thats the other half , 88 errors worth , same shape of work

@HenryNdubuaku on that , do you want it folded into this pr , a second one stacked on top , or would you rather not carry a slow path at all. either way this pr changes no build flags , so nothing moves for chips that have dotprod , and i have the pi here to measure whichever way you prefer

